PR #1, opened 2022-03-28
add: support `straight`

theFool32:
add `provide` to support package manager.
Also, add a configuration example for `straight`.


zerolfx:
Thanks for contributing! Looks good to me.

PR #24, opened 2022-06-08
Add `copilot-mode-map`

SqrtMinusOne:
`copilot-mode` doesn't ship with a mode map.

Defining `copilot-mode-map` before `copilot-mode` should allow for defining user keybindings in this keymap. I've added an example to the README.


zerolfx:
Thank you for contributing!

PR #27, opened 2022-06-15
Add integration with lispy

SqrtMinusOne:
I'm using abo-abo's [lispy](https://github.com/abo-abo/lispy) to edit Lisps. Among other things, the package does a pretty good job of preserving the balance of parentheses, i.e. using its operations (like deleting a line) should never break the balance.

And accepting a suggestion from copilot seems to break the balance in most cases.

So this PR adds an option (`copilot-lispy-integration`) that changes two things:
- using `lispy--maybe-safe-delete-region` instead of `delete-region` to avoid deleting unmatched parenthesis
- processing the completion with `lispy--find-safe-regions` to avoid inserting unmatched parenthesis.


zerolfx:
From my experience of editing elisp, Copilot preserves the balance of the parentheses in ~2/3 of cases, which is not good enough but acceptable.

`lispy` looks like an excellent package. I will try it as well as your PR and give you feedback later.

Is it a good idea to simply extract a function named `copilot-insert-completion` so you can use `advice-add`? Because I want to make `copilot.el` a general package, and the balance problem also exists in other languages.

SqrtMinusOne:
In that case I'd opt for `run-hook-with-args-until-success`. It's easier to change a list of hooks than to advice a function.

SqrtMinusOne:
Something like the latest commit should do this.

zerolfx:
I find the lispy integration doesn't work well with `copilot-accept-completion-by-word`. It accidentally deletes some prefix parentheses and quotes.

SqrtMinusOne:
Yeah, it seems so.

But the problem is that in the lispy paradigm there cannot be an operation that leaves an unbalanced parenthesis. So, if we want to insert `(hello`, there are two valid ways:
- `(hello)`
- `hello`

I.e. the in general case we either insert missing parentheses or delete extra ones. I tried both and the latter seem to backfire way less often (also, lispy has some limitations with regards to quotes).

It should be possible to make a more "intelligent" decision. But I have little idea on how to approach this and if it's even worth the effort, because lispy itself fails every now and then (and it's ~16K LoC in total).

PR #31, opened 2022-06-21
Add a global mode and a keymap for the completion overlay

mpontus:
Hey @zerolfx. Amazing work on this plugin. After using VSCode Copilot extension I was very impressed by how easy it was to achieve a similar experience in Emacs as a result of your efforts.

As a token of appreciation, I'd like to propose following changes, which I think could improve this experience even more:

- Define `global-copilot-mode` to enable `copilot-mode` in all buffers, which could be convenient if you want to use copilot suggestions for text as well as code.
- Define `copilot-suggestion-map`, which becomes active while suggestion overlay is shown, changing overlay's `FRONT-ADVANCE` arg to nil in order for the point to be inside overlay boundaries.

By default it binds **`TAB`** to `copilot-accept-completion`, but it can also be used to bind other completion commands, e.g.:

```elisp
(use-package! copilot
  :config (global-copilot-mode)
  :bind (:map copilot-completion-map
         ("RET" . #'copilot-accept-completion)
         ("<tab>" . #'copilot-accept-completion-by-line)
         ("C-<tab>" . #'copilot-accept-completion-by-word)
         ("C-n" . #'copilot-next-completion)
         ("C-p" . #'copilot-previous-completion)))
```

### DIFF
--- a/copilot.el
+++ b/copilot.el
@@ -478,6 +478,10 @@ Use this for custom bindings in `copilot-mode'.")
   (advice-add 'posn-at-point :before-until 'copilot--posn-advice)
   (add-hook 'post-command-hook 'copilot--complete-post-command))
 
+;;;###autoload
+(define-global-minor-mode global-copilot-mode
+    copilot-mode copilot-mode)
+
 (defun copilot--complete-post-command ()
   "Complete in post-command hook."
   (when (and copilot-mode this-command)
--- a/copilot.el
+++ b/copilot.el
@@ -322,6 +322,11 @@ To work around posn problems with after-string property.")
                (eq pos (car copilot--real-posn)))
       (cdr copilot--real-posn))))
 
+(defconst copilot-completion-map (make-sparse-keymap)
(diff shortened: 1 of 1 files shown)


zerolfx:
For `copilot-completion-map`, what will happen if:

1. a user uses `company` for completion and uses tab key to complete
2. copilot and company show completions at the same time
3. user hits tab

Users may sometimes fail to complete company (or copilot) by tab. Will it be confusing?

Also, it is a breaking change, while some users use tab for other purposes (and bind different keys to copilot).

Though I think tab is a good default, leaving the map empty seems a safer choice.

mpontus:
@zerolfx agreed and done

zerolfx:
Thanks for your work!

anildigital:
@mpontus Thanks for the great work! Which keybinding do you hit when you want to ignore a copilot suggestion? Looks like sometimes I get stuck as I can't come out of suggestion loop.

PR #32, opened 2022-06-22
Fix byte-compile error

syohex:
Functions which are used in macros should be evaluated at both compile time and load time

```
copilot.el:138:4: Error: Symbol’s function definition is void: copilot--transform-pattern
```


zerolfx:
I noticed the compile error before but didn't know how to fix it.

Thanks for your PR!!

Issue #23, opened 2022-06-05
Unable to login


ladrua:
When I first start emacs and try to login I get error: `No applicable method: jsonrpc--request-continuations, nil`

Then I run `copilot-diagnose` and get "NotSignedIn" error. Then if I retry `copilot-login` I get: `jsonrpc-error: "request id=4 failed:", (jsonrpc-error-code . -32602), (jsonrpc-error-message . "Invalid params:  must be object"), (jsonrpc-error-data)`

Copilot-events:
```
[client-request] (id:1) Sun Jun  5 23:38:04 2022:
      (:jsonrpc "2.0" :id 1 :method "initialize" :params
                (:capabilities 'nil))
      [server-notification] Sun Jun  5 23:38:05 2022:
      (:jsonrpc "2.0" :method "LogMessage" :params
                (:level 0 :message "[DEBUG] [agent] [2022-06-05T21:38:05.280Z] Agent service starting" :metadataStr "[DEBUG] [agent] [2022-06-05T21:38:05.280Z]" :extra
                        ["Agent service starting"]))
      [server-reply] (id:1) Sun Jun  5 23:38:05 2022:
      (:jsonrpc "2.0" :id 1 :result
                (:capabilities
                 (:textDocumentSync
                  (:openClose t :change 2)
                  :workspace
                  (:workspaceFolders
                   (:supported :json-false :changeNotifications :json-false)))))
      [client-request] (id:2) Sun Jun  5 23:38:05 2022:
      (:jsonrpc "2.0" :id 2 :method "setEditorInfo" :params
                (:editorInfo
                 (:name "Emacs" :version "25.2.2")
                 :editorPluginInfo
                 (:name "copilot.el" :version "0.9.2")))
      [client-request] (id:3) Sun Jun  5 23:38:05 2022:
      (:jsonrpc "2.0" :id 3 :method "getCompletions" :params
                (:doc
                 (:source "\n" :path "" :uri "" :relativePath "" :languageId "text" :position
                          (:line 0 :character 0))))
      [server-reply] (id:2) Sun Jun  5 23:38:05 2022:
      (:jsonrpc "2.0" :id 2 :result "OK")
      [server-reply] (id:3) ERROR Sun Jun  5 23:38:05 2022:
      (:jsonrpc "2.0" :id 3 :error
                (:code 1000 :message "Not authenticated: NotSignedIn"))
      [client-request] (id:4) Sun Jun  5 23:38:11 2022:
[...]

PR #45, opened 2022-08-08
Raise error if there's an issue with node

ethan-leba:
This prevents the issue of the agent start implicitly failing to set up
the connection and leading to hard to understand error with jsonrpc. See
https://github.com/zerolfx/copilot.el/issues/23#issuecomment-1207579974


zerolfx:
LGTM. Thank you!

PR #47, opened 2022-08-09
copilot-mode: use after-change-functions instead of post-command

kiennq:
Thanks for your great package.
In this change, I propose to use the `after-change-functions` instead of `post-command-hook` to make the copilot request only when the document has been changed.
Also minor change in minor mode activation to correctly add/remove the function from the local hooks. For the `posn-at-point` function, guarding using `copilot-mode` instead as this is advice.
Finally, debounce the whole hook function using idle timer instead of only the request, this should make it less interrupt for the user's typing


zerolfx:
I use `post-command` because the overlay needs to be clear after operations like pressing Esc (for evil users) or moving the cursor. I just tried this patch, and this problem does exist.

But I love these minor changes. Thanks for contributing!

kiennq:
Thanks, I've changed back to using `post-command-hook` and also update the logic to detect buffer changes based on the last time `copilot-complete` has been invoked.

zerolfx:
Thanks again for your help.

VitalyAnkh:
Should the commit message be updated to reflect the actual changes?

zerolfx:
> Should the commit message be updated to reflect the actual changes?

Thank you! Now the commit message is fixed.

PR #48, opened 2022-08-09
copilot--overlay: reuse when possible

kiennq:
As the title, when we delete the overlay, it will be piled up and wait for being removed by the garbage collector. We should try to reuse the overlay instead.

### DIFF
--- a/copilot.el
+++ b/copilot.el
@@ -286,16 +286,21 @@
                  (let ((completion (elt completions idx)))
                    (copilot--show-completion completion))))))))
 
+(defsubst copilot--overlay-visible ()
+  "Return whether the `copilot--overlay' is avaiable."
+  (and (overlayp copilot--overlay)
+       (overlay-buffer copilot--overlay)))
+
 (defun copilot-next-completion ()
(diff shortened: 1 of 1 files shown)


zerolfx:
I guess that `move-overlay` is seldomly triggered.
The completion steps are: user input => clear overlay => wait for copilot response (~100ms) => show completion
So the overlay is deleted before showing the next completion.

Maybe in `copilot-clear-overlay`, instead of calling `delete-overlay`, making it invisible would be better?

Anyway, reducing GC time is a good idea; I had never thought about it before.

kiennq:
My understanding is that `delete-overlay` is simply removing the overlay from its buffer, whereas the `make-overlay` is used to allocate a completely new overlay.
The `move-overlay` will move the overlay which has been stripped from the current buffer back to it, so basically we're reusing one overlay instead of creating a new one every time showing the completion response

zerolfx:
I thought `delete-overlay` would cause `overlayp` to return `nil`. Sorry for my confusing comment.

PR #55, opened 2022-09-14
add: support a list of commands that should be allowed to call without clearing the copilot overlay

hrehfeld:


hrehfeld:
(Documentation should be updated, but I don't really have time right now)

zerolfx:
LGTM!

PR #71, opened 2022-12-27
Add requirement `use-package`

philippludwig:
It seems that `use-package` is required for this to work, at least for me this was the case.


zerolfx:
The `use-package` package provides a single macro `use-package`, but it is not used in the example configuration that you updated.

As I mentioned in the readme, you can also use `load-file` or `require` instead of `use-package`.

philippludwig:
That part in the README applies to loading it manually, but my change was about when you use `straight.el`.

PR #74, opened 2023-01-05
Add `python-ts-mode` to indentation alist

nasyxx:


nasyxx:
> Which package contains `python-ts-mode`? 

python-ts is the new treesit python mode in Emacs. See lisp/progmodes/python.el

> Is `python-ts-mode` derived from `python-mode`?

No, it is derived from `python-base-mode`.

zerolfx:
You may also create a PR for https://github.com/editorconfig/editorconfig-emacs.

PR #88, opened 2023-02-20
update agent.js to version of...

rksm:
...https://github.com/github/copilot.vim/tree/release/copilot/dist

This is just taking the latest version from copilot.vim and fixing an error regarding the missing version property. Used it a bit and as far as I can tell completions are working fine but there might be potentially changes in the interface that are not covered.

Do you know if there is any documentation / changelog for the VSCode extension? Seems really annoying that they just ship that as a blob.


zerolfx:
I don't know any documentation for the VSCode extension, but the VSCode version does have more functionalities.

I plan to upgrade agent.js after syncing the changes from https://github.com/github/copilot.vim/commit/8c8416488ef21483dd5d2922ea194c05e9a3baaf (and when I have enough time). I would appreciate it if you could help me with that.

rksm:
I mean I'm not familiar with the vim plugin but there are probably a few areas of improvement.

Not sure if the language id really matters, it is currently derived from the mode name. The vim plugin has a map with a few ids: https://github.com/github/copilot.vim/blob/release/autoload/copilot/doc.vim#L14. Pretty printing agent.js and looking into it, it seems the full list would be:
```
abap
bat
bibtex
blade
c
clojure
coffeescript
cpp
csharp
css
dart
dockerfile
elixir
erb
erlang
fsharp
go
groovy
haml
handlebars
haskell
html
ini
java
javascript
javascriptreact
jsonc
jsx
julia
kotlin
latex
less
lua
makefile
markdown
objective-c
objective-cpp
perl
php
powershell
pug
python
ql
r
razor
ruby
rust
sass
scala
scss
shellscript
slim
solidity
sql
stylus
svelte
swift
terraform
tex
typescript
typescriptreact
vb
verilog
vue-html
vue
xml
xsl
yaml
php
plaintext
```
So making a mapping to ensure those "ids" are used is something that could be improved. Whether or not this makes a difference regarding the completion quality I have no idea.

As for the document `version` property that now seems to be required: It seems the vim plugin sends `textDocument/didChange` notification to some lsp server and then uses the version tracking of the lsp client/server as the version property to send to the copilot agent.
Not sure what's the lsp endpoint in this case. It isn't the agent I think, I tried sending that notification but just got errors. In any case, just using 0 as version seems to work OK, the agent does not increment its internal version apparently.

Other than that there don't seem to be much more changes.

NotBrianZach:
can confirm this pull request fixes copilot plugin for me after it refused to start after I updated nodejs to version 18.

PR #93, opened 2023-02-21
feat: add dwim functions which trigger or accept existing completion

jimeh:
I find this behavior useful, either after having manually dismissed a completion with C-g, or on rare occasion when Copilot doesn't offer any completions, but forcing it to with copilot-complete does bring something up.

This is based on my [personal config](https://github.com/jimeh/.emacs.d/blob/a8e92eb311768e52d8452c7b10610c4b8eafe0ac/modules/completion/siren-copilot.el), where I use `<backtab>` (shift+tab) to both force-trigger and also accept Copilot completions. This leaves `<tab>` free to work like normal with lsp-mode/company-mode/etc.

### DIFF
--- a/copilot.el
+++ b/copilot.el
@@ -508,6 +508,31 @@ Use TRANSFORM-FN to transform completion if provided."
               (message "No completion is available."))))))))
 
 ;;
+;; dwim - do what I mean
+;;
+
+(defun copilot-dwim ()
+  "Trigger or accept completion."
+  (interactive)
+  (if (copilot--overlay-visible)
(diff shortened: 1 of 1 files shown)


zerolfx:
Sorry, I can't merge this PR, but thanks for your suggestion.
This project aims to provide an auto-complete experience similar to VSCode while making it easy to customize/hack.
So I can't incorporate all user flavors but only provide a standard one.

jimeh:
That makes perfect, thanks for taking the time to review :)

PR #113, opened 2023-03-29
Add function to accept completion by paragraph

debanjum:
- Add helper function to accept completion by paragraph
  This is similar to the existing `accept-completion-by-word`, `accept-completion-by-line`
- Make `accept-completion-by-word`, `accept-completion-by-line` function arguments optional
  It already defaults to 1 when the `n-words`, `n-lines` args are `nil`


ntc2:
This function isn't documented in the README.

PR #116, opened 2023-03-30
support middle-of-the-line completions

Rogach:
GitHub Copilot usually returns no completions if position is not at the end of the line. However if position is only one character from the end completions are sometimes returned, but the behavior is unreliable - sometimes the last character is returned as part of completion, sometimes not.

This commit attempts to handle this case by removing common suffix between completion and current line prior to insertion.

I also converted overlay code to use zero-width overlays at all times - in order to unify the code paths and get rid of special cases for eof and eol.

For some reason `(put-text-property 0 1 'cursor t)` didn't work as expected - so I had to replace it with `(put-text-property 0 1 'cursor 1)` for now to get correct cursor positioning if completion is in the middle of the line. I opened a [bug report](https://debbugs.gnu.org/cgi/bugreport.cgi?bug=62540) to Emacs in order to clarify the issue.

### DIFF
--- a/copilot.el
+++ b/copilot.el
@@ -392,17 +392,13 @@ To work around posn problems with after-string property.")
 
 (defun copilot--set-overlay-text (ov completion)
   "Set overlay OV with COMPLETION."
-  (move-overlay ov (point) (line-end-position))
+  (move-overlay ov (point) (point))
   (let ((p-completion (propertize completion 'face 'copilot-overlay-face)))
-    (if (eolp)
-        (progn
-          (overlay-put ov 'after-string "") ; make sure posn is correct
-          (setq copilot--real-posn (cons (point) (posn-at-point)))
-          (put-text-property 0 1 'cursor t p-completion)
-          (overlay-put ov 'display "")
-          (overlay-put ov 'after-string p-completion))
-      (overlay-put ov 'display (substring p-completion 0 1))
-      (overlay-put ov 'after-string (substring p-completion 1)))
+    (overlay-put ov 'after-string "") ; make sure posn is correct
+    (setq copilot--real-posn (cons (point) (posn-at-point)))
+    (put-text-property 0 1 'cursor t p-completion)
+    (overlay-put ov 'display "")
+    (overlay-put ov 'after-string p-completion)
     (overlay-put ov 'completion completion)
     (overlay-put ov 'start (point))))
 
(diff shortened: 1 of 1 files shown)


Rogach:
Never saw Copilot offering multi-line completion in the middle of the line, but I tested it now by hardcoding the completion inside `copilot--show-completion` - seems to work fine for me, for all cases I managed to think of.

zerolfx:
Before `copilot-complete`:
image
After `copilot-complete`:
image

In my opinion, it's necessary to hide the remaining part of the current line using `(move-overlay ov (point) (line-end-position))`.

zerolfx:
At the beginning of the project, after struggling with overlays and the special case (at the end of lines), I referenced code from `auto-complete`.

https://github.com/auto-complete/auto-complete/blob/59c11c453bfa0794302848a004d1be3b1684f17b/auto-complete.el#L928-L937

Rogach:
Yes, for python and multiline completion removing the rest of the line seems to make sense. Maybe we can hide the rest of the line when completion is multiline, but keep it when it's single-line?

For reference, here's the problem I'm talking about:

Before `copilot-complete`:
copilot el-before
After `copilot-complete`:
copilot el-after
After `copilot-accept-completion`:
copilot el-accept

(note the hilarious trick of asking Copilot to generate the buggy completion :)

Here's the source:

```python
pyplot.stairs(x, ys + [], fill=True, color="tab:green")
# the next line completion often has no closing brace at the end
pyplot.stairs(x, ys + [], fill=True)
```

<hr>

I'm feeling uneasy about hiding the rest of the line, because this hides "my" input and that's sometimes quite sneaky, especially if I'm typing something at the same time (I have to first dismiss copilot completion to check if there's something inside or not).

Maybe the optimal solution would be to the remainder of the current line if the completion is multi-line, but keep the remainder visible if completion is only single-line?

zerolfx:
There are more subtle cases:

Before `copilot-complete`:
image
After `copilot-complete`:
image
After `copilot-complete` (with this PR):
image
CLion's behavior:
image


I believe that Copilot's "hiding" behavior is reasonable, and usually only affects parentheses. If you add random letters at the end of a line, Copilot won't provide multiline completions which hide those letters. 

Personally, I think it would be best to maintain the current behavior and keep it consistent with other IDEs such as VSCode or JetBrains IDEs.

PR #124, opened 2023-04-11
Suggest setting `copilot-enable-predicates` to nil for spacemacs

### LINKED ISSUE #123: copilot-complete not automatic
I have copy and pasted exactly what is in the spacemacs configuration part of the [README](https://github.com/zerolfx/copilot.el#example-for-spacemacs) and looking at the [General configurations for configuring completetion](https://github.com/zerolfx/copilot.el#option-1-use-copilot-mode-to-automatically-provide-completions) all I should need is `(add-hook 'prog-mode-hook 'copilot-mode)`, which I have in my user-config. 

However, unless I explicitly run the command `copilot-complete`, I'm not getting automatic suggestions, any help would be greatly appreciated.
Screencast from 06-04-23 16 42 22

james-ennis:
In order for this plugin to work with spacemacs we need to set `copilot-enable-predicates` to `nil`. This closes #123

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -94,6 +94,8 @@ dotspacemacs-additional-packages
 
 (define-key evil-insert-state-map (kbd "C-<tab>") 'copilot-accept-completion-by-word)
 (define-key evil-insert-state-map (kbd "C-TAB") 'copilot-accept-completion-by-word)
+
+(setq copilot-enable-predicates nil)
 ```
 
 </details>


zerolfx:
The `copilot-enable-predicates` contains two functions by default. Can you please confirm which one is causing the plugin to not work with Spacemacs?

james-ennis:
I tried setting it to both `copilot--buffer-changed` and  `evil-insert-state-p` individually, neither worked. 

These were the two functions I found when I did `M-x describe-varable copilot-enable-predicates`

zerolfx:
`(setq copilot-enable-predicates '(evil-insert-state-p))`
`(setq copilot-enable-predicates '(copilot--buffer-changed))`

In case you forgot to make it a list.

james-ennis:
Ah yes, I didn't make them a list. Thanks, so it's:
```
(setq copilot-enable-predicates '(evil-insert-state-p))
```
that is _preventing_ the plugin from working. I wonder if this is because I use spacemacs with emacs keybindings rather than vim ones. That setting sounds very vim-like

zerolfx:
Yes, `evil-insert-state-p` will prevent the plugin from triggering completions if the current VIM state is not `insert`. However, `copilot.el` will ignore this function if it's not available, so it should work for users using Emacs keybindings.

I'm curious why `(functionp evil-insert-state-p)` returns `t`, even when you're using Emacs keybindings.

james-ennis:
I'm not actually able to see `functionp` when I run `M-x`, not to sure why. I can describe it though: 
```
evil-insert-state-p is a native-compiled Lisp function in ‘hybrid-mode.el’.

(evil-insert-state-p &optional STATE)

Whether the current state is insert.
```
So what shall we do with PR? It sounds to me as though you're suggesting it could be an issue with Spacemacs?

zerolfx:
Spacemacs offers three editing styles: Vim, Emacs, and Hybrid. However, `hybrid-mode` is beyond the scope of consideration due to the small number of users.

I recommend that you switch your editing style from hybrid to emacs, or alternatively customize `copilot-enable-predicates`.

urbanslug:
Hello @zerolfx I am using emacs keybindings and I had to set `(setq-default copilot-enable-predicates nil)` in my `.spacemacs` config for copilot to work as expected.

fnussbaum:
I am going to fix this in Spacemacs.

PR #127, opened 2023-04-18
Update readme to respect minibuffer on tab completion

JeffAbrahamson:


zerolfx:
The preferred way to add keybindings has been updated to use `copilot-completion-map`, which will only activate with the copilot overlay. The "General Configurations" section of README will be updated at a later time.

PR #130, opened 2023-04-22
Allow disabling SSL cert verification when using a proxy

### LINKED ISSUE #129: Allow disabling SSL when using a proxy
copilot.vim has an option, `g:copilot_proxy_strict_ssl`, which can be set to `false` to disable SSL verification when using a proxy. For a MITM proxy, this is basically required. It'd be great to have an equivalent feature added to copilot.el.

spenczar:
Fixes #129.

I had to write a helper function to get the right json encoding, since we need a JSON `false` for `rejectUnauthorized`.

I haven't written much lisp so let me know if the copilot--proxy-param is a horrible way to do this. It sure feels ugly.

### DIFF
--- a/copilot.el
+++ b/copilot.el
@@ -23,7 +23,14 @@
 (defcustom copilot-network-proxy nil
   "Network proxy to use for Copilot. Nil means no proxy.
 Format: '(:host \"127.0.0.1\" :port 80 :username \"username\" :password \"password\")
-Username and password are optional."
+Username and password are optional.
+
+If you are using a MITM proxy which intercepts TLS connections, you may need to disable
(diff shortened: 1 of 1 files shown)


zerolfx:
How about updating the documentation and instructing users to add `:rejectUnauthorized :json-false` if needed? The only limitation is that this sub-option cannot be configured through the UI.

spenczar:
Oh, that makes good sense to me. So users would just setq it themselves pretty explicitly - yeah, that seems much nicer. I will revise this when I get back to a computer in a few hours.

spenczar:
This is ready for another look. Much nicer!

Issue #117, opened 2023-04-02
Completion removes the rest of the line


w9:
If we start with the following text, where "`|`" is the cursor:

    ("|")

and start typing:

    ("H|")

The completion will delete the the rest of the line starting from the cursor, in this case the `")`

    ("H|ello, World!

This is highly undesirable especially when smartparens or lispy mode are used, and the parentheses balance are ruined.

zerolfx:
The issue appears to be a result of Copilot, and unfortunately I have not been able to find a solution within copilot.el itself.

See also #116 #27 https://github.com/orgs/community/discussions/50038 https://github.com/orgs/community/discussions/46277

Issue #77, opened 2023-01-07
copilot clobbers closing parentheses when there is only 1 character left


LemonBreezes:
Hello. With the current code, if I am writing Lisp code, copilot offers an autocompletion of 't' and I type the character 't', the rest of the line including closing delimiters is deleted. It would be helpful if we could turn off this "autocompletion of the last character" behavior.

zerolfx:
Could you provide a detailed example, like what is on your screen before and after typing 't'? It would be better if you could record an GIF.

LemonBreezes:
> Could you provide a detailed example, like what is on your screen before and after typing 't'? It would be better if you could record an GIF.

Sure. I just recorded a rudimentary example with my phone (not sure how to make a GIF).https://user-images.githubusercontent.com/29102529/211169968-23de18b6-c32f-4870-9a4d-e98611ff734e.mp4

zerolfx:
Fixed.  The issue was introduced in #73.

Issue #62, opened 2022-11-02
copilot.el clobbering parens with suggestions


jjtolton:
First of all -- AMAZING!  If this is something I can dive in with, happy to if you can point me in the right direction.

The issue is effectively that autocomplete is clobbering parenthesis.  Please see the GIF.  

![clobbered](https://user-images.githubusercontent.com/10394611/199578166-6a57fe7f-90b3-40ae-8c87-d112ab922c0c.gif)

It's a minor nuisance but it does break stride.

jjtolton:
Ok still working on this a bit because it's tricky but I have something that for lisp-languages at least makes it paredit/lispy safe (meaning, the parens will be BALANCED after insertion -- not necessarily perfect) but it won't ruin your life.

![beforeFix](https://user-images.githubusercontent.com/10394611/200082804-a901a7c3-861f-4f6c-9632-213892fef1cb.gif)
![afterFix](https://user-images.githubusercontent.com/10394611/200082805-1d33f96b-14d9-40de-90f1-e17acabe85e7.gif)

No external dependencies.  Funny enough I have no idea how to write elisp, but I `copilot.el` helped me write it! (this also takes into account double quoted strings and skips over them)  :1st_place_medal:  Great stuff!

Would love to start a conversation about whether or not this is a good experience or if I should just keep it local. 

I was looking into some more sophisticated libraries to see how the approach alternate grammars and it looks like quite a lot of work: 
![image](https://user-images.githubusercontent.com/10394611/200083062-6b9082d7-7940-4e78-aed6-d35798da819a.png)

This is an example from the [`expand-region.el`](https://github.com/magnars/expand-region.el) library which handles syntax highlight for a number of grammars, and it looks pretty modular.  I don't know how much of an issue this outside of lispy languages where breaking the parenthetical grouping is devastating to the structural editing workflow.  

For the record, I tested this on JetBrains's copilot plugin, and they have the same paren clobbering issue that `copilot.el` does.

zerolfx:
See #27. It's a known issue based on the nature of Copilot, and we don't have a perfect solution yet.

jjtolton:
My fix works for me, so I'm closing this.  Thanks again for this amazing, amazing tool.

PR #132, opened 2023-04-22
Fix handling of completions when they don't go to the end of line

CeleritasCelery:
Currently copilot.el assumes that the completion text will go all the way to the end of the line, but this is not always the case. When this happens it will lead to incorrectly deleting the text that used to be there. This will lead to unbalanced parens or quotes.

This PR fixes this by using the API's `:end` property to determine where the end of the completed text should go.

Example:
text given to copilot
```python
print("hello |")
```
copilot will return the text `world`.

With the current implementation this will be the result if the user accepts the completion:
```python
print("hello world
```

After this PR, the result would be as expected:
```python
print("hello world")
```

This will resolve issues like #77 #62 #117

### DIFF
--- a/copilot.el
+++ b/copilot.el
@@ -418,8 +418,7 @@ To work around posn problems with after-string property.")
 (defun copilot--set-overlay-text (ov completion)
   "Set overlay OV with COMPLETION."
   (move-overlay ov (point) (line-end-position))
-  (let* ((line-end (line-end-position))
-         (tail (buffer-substring (- line-end (overlay-get ov 'end)) line-end))
+  (let* ((tail (buffer-substring (overlay-get ov 'end) (line-end-position)))
          (p-completion (concat (propertize completion 'face 'copilot-overlay-face)
                                tail)))
     (if (eolp)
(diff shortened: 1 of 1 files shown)


zerolfx:
Thanks for your work. You have solved a long-standing problem.

I made some minor changes by refactoring the code and changing the meaning of the `end` property from length to absolute position. Additionally, I removed the calls to `goto-line`, as it is only intended for interactive use.

CeleritasCelery:
Thanks. I initially tried to use end as an absolute position like you did, but it doesn't work when you have automatic pair insertion (via smartparens, electric-pair-mode, cleverparens, etc). The issue is that if I insert `(` it will also insert `)`. The end position should move forward 2 places, but we only advance it one with `cl-incf`. I don't know of a reliable way to detect the auto-pair case, so that is why I defined `end` to be an offset from the end of the line.

zerolfx:
Oh no, I haven't tested that case. I will revert my changes if there is a bug.

zerolfx:
Fixed in https://github.com/zerolfx/copilot.el/commit/7986e38726879350c2d2a13e4751e99f1a89daed

PR #135, opened 2023-05-02
Add get panel completions support

rwc9u:
Add support for calling `getPanelCompletions`. 

In neovim this is called with `:Copilot panel` or `:Copilot`
> Open a window with up to 10 completions for the current buffer.

### DIFF
--- a/copilot.el
+++ b/copilot.el
@@ -9,6 +9,7 @@
 (require 's)
 (require 'dash)
 (require 'editorconfig)
+(require 'org)
 
 (defgroup copilot nil
   "Copilot."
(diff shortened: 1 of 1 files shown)


zerolfx:
I have made some changes to the code. I attempted to use `org-mode` for displaying completions as line separators can cause syntax errors in certain languages and disrupt syntax highlighting. However, I encountered an org-related error: `Error running timer ‘org-reveal’: (wrong-type-argument number-or-marker-p nil)`. I will address this issue later.

To make the panel feature better, we can also:
- sort the results by score
- filter out duplicated results

rwc9u:
Cool. I'll try those others. I'm a relative newb at programming more than a few lines of emacs-lisp.

rwc9u:
If you would prefer that I stack commits in separate PRs I can do that. I didn't see the error you mentioned with my `org-mode` setup. I have added sorting based on score. I did the sorting for each `PanelSolution` notification because sometimes I don't get 10 solutions so I never get the `PanelSolutionsDone`.

zerolfx:
Thank a lot for your effort.

You made excellent use of org-mode properties to sort and filter. (I'm not an org user.)

There is one minor issue that remains: `mark-whole-buffer` should not be used in elisp programming because it is interactive-only. Unfortunately, we cannot call `org-sort-entries` without using it. I will merge the changes now and address the warning later if necessary.

PR #146, opened 2023-06-04
Enable workspace capabilites

haukot:
They are enabled in Copilot.vim https://github.com/github/copilot.vim/blob/1358e8e45ecedc53daf971924a0541ddf6224faf/autoload/copilot/agent.vim#L536


danp:
Just curious: what are the implications of this change?

PR #149, opened 2023-06-14
Allow enabling copilot only in specific major modes

solita-perttikellomaki:
In a corporate environment it is nice to be able to control explicitly in which modes copilot is turned on.

The selection is done in .emacs like this:

  (setq copilot-only-in-modes
        '(clojure-mode clojurescript-mode))

### DIFF
--- a/copilot.el
+++ b/copilot.el
@@ -62,6 +62,11 @@ Enabling event logging may slightly affect performance."
   :group 'copilot
   :type '(repeat function))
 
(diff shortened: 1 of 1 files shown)


zerolfx:
Instead of using `global-copilot-mode`, why not use `copilot-mode`?

You can use something like `(add-hook 'clojure-mode-hook #'copilot-mode)`.

PR #157, opened 2023-07-13
[RFC] add copilot-balancer.el to fix unbalanced char pairs returned from Copilot

raymond-w-ko:
One thing that has frustrated me greatly is that Copilot cannot balance parentheses et al, even when using VSCode + the official plugin, which is the golden standard. Since I program in Clojure for work, this is maddening as you have to these fix missing or extra char pairs. Almost all lisp editors run into issues once in this unbalanced state.

Thankfully, the brokenness follows a pattern where I think it is only the suffix portion that is mismatched, meaning it can be algorithmically fixed. I attempt to do so here.

There are other two changes:
1. I coerce all Clojure modes to just `clojure` as that is the only official language ID.
2. I rewrote the code so that `save-excursion` is in the outer loop and `save-restriction` is inside, as recommended by the docs. I also eliminated multiple calls as a micro-optimization.

### DIFF
--- a/copilot-balancer.el
+++ b/copilot-balancer.el
@@ -1,0 +1,216 @@
+;; -*- lexical-binding: t -*-
+
+(require 'dash)
+(require 'rx)
+
+(defvar copilot-balancer-lisp-modes '(emacs-lisp-mode
+                                      lisp-mode
+                                      scheme-mode
+                                      clojure-mode)
+  "List of lisp modes to balance.")
+
+(defvar copilot-balancer-lisp-pairs
+  (let ((h (make-hash-table :test 'equal :size 7)))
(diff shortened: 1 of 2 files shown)


zerolfx:
LGTM! Great work!

I can't push to your branch, please resolve the conflict and I'll merge it soon.

raymond-w-ko:
There is still one bug I'm working on, so let me see if I can fix it:

```emacs
(defun add () | 
  )(EOF)
```

raymond-w-ko:
I had to rewrite portions to deal with pairs inside strings, so I'm not 100% as I haven't used it for my work. However, it should be ready to merge.

Issue #150, opened 2023-06-19
Differences from JetBrains version


haukot:
HI! I've done research and reverse-engineered the Github Copilot protocol between IDE plugin and agent.js, and have tried to document it. This is based on Github Copilot plugin version for JetBrains and some parts of agent.js (repo https://github.com/haukot/copilot_specifications)

I'd like to make Emacs plugin a bit more complete. Specifically I want to add commands "textDocument/didFocus"(to have better context based on last access time), "textDocument/didChange" with range(so it'll not send whole document to the agent.js on each request), and "textDocument/didClose"(to remove unused files from current context).
And probably update agent.js binary.

Will you consider PR with these changes if I made them, or do you want to stick with the Vim implementation?

zerolfx:
One crucial missing feature is the ability to provide completions based on other files. 

Thank you for creating these wonderful specifications, which are much easier to read than Copilot.vim's source code. 
I truly appreciate your efforts in bridging the gap between Copilot in Emacs and Jetbrains.

There are other projects that will benefit from your research, such as:
- https://github.com/zbirenbaum/copilot.lua
- https://github.com/TerminalFi/LSP-copilot

haukot:
Thanks!

>There are other projects that will benefit from your research, such as:

Thanks for the note, I've sent them too!

>One crucial missing feature is the ability to provide completions based on other files.

It actually already exists and works, it just could work better.

It works on the mechanic based on 'open tabs' (or, in the emacs case, open buffers). 
When you open a file(with didOpen), the file will be added to the context with files of the same language.
But there are limitations - only 20 files max could be in the context, and not all content from these files is used(because there are limitations to the prompt length). 
So all files that you open are firstly sorted by the time of the last access, and after that from the file parts the prompt is built(the prompt creating algorithm is actually pretty complex and I don't know much about it. There are other analyses about that)

And two things, that Vim implementation doesn't have, are an event on a file focus("textDocument/didFocus", which sets the last access time),  and an event on a file close("textDocument/didClose", which removes the file from context)

philipsd6:
Another item we would like is the ability to report telemetry, so we can gather insights on number of code completions accepted, and number accepted by language. VSCode does that but I want to keep using Emacs. Obviously this would have to be an opt in feature.

haukot:
[...]

PR #164, opened 2023-08-09
[improvement] Improve active buffer context

emil-vdw:
Implement the following suggestions from #150:
- `textDocument/didFocus` on focus of already opened buffer (`window-selection-change-functions` hook)
- `textDocument/didOpen` on first focus of buffer (`window-selection-change-functions` hook)
- `textDocument/didClose` event on buffer kill (`kill-buffer-hook` hook)
- `textDocument/didChange` event on buffer change (`before-change-functions` hook for deletions and `after-change-functions` for insertions) with only data relevant to the change not the whole buffer.

### DIFF
--- a/copilot.el
+++ b/copilot.el
@@ -625,43 +625,36 @@ Use TRANSFORM-FN to transform completion if provided."
   ;; send a notification for the window gaining focus and only if the buffer has
   ;; copilot-mode enabled.
   (when (and copilot-mode (eq window (selected-window)))
-	(if (-contains-p copilot--opened-buffers (current-buffer))
-		(progn
-          (copilot--notify ':textDocument/didFocus
-						   (list :textDocument (list :uri (copilot--get-uri)))))
+    (if (-contains-p copilot--opened-buffers (current-buffer))
+        (copilot--notify ':textDocument/didFocus
+                         (list :textDocument (list :uri (copilot--get-uri))))
       (add-to-list 'copilot--opened-buffers (current-buffer))
       (copilot--notify ':textDocument/didOpen
                        (list :textDocument (list :uri (copilot--get-uri)
-												 :languageId (copilot--get-language-id)
-												 :version copilot--doc-version
-												 :text (copilot--get-source)))))))
+                                                 :languageId (copilot--get-language-id)
+                                                 :version copilot--doc-version
+                                                 :text (copilot--get-source)))))))
 
-(defun copilot--on-doc-change (&optional start end chars-replaced-length)
+(defun copilot--on-doc-change (&optional beg end chars-replaced)
   "Notify that the document has changed."
-  (let* ((is-after-change (not (eq chars-replaced-length nil)))
-         (is-before-change (not is-after-change))
-         (is-insertion (or (and is-before-change (eq start end))
-                           (and is-after-change (eq chars-replaced-length 0))))
-         (is-deletion (not is-insertion)))
-
-    (when (or (and is-before-change is-deletion)
-			  (and is-after-change is-insertion))
-      (let* ((text (if is-insertion (buffer-substring-no-properties start end) ""))
-             (range-start (list :line (- (line-number-at-pos start) copilot--line-bias)
-                                :character (- start (save-excursion (goto-char start) (line-beginning-position)))))
-
-             (range-end (if is-insertion
-                            range-start
+  (let* ((is-before-change (eq chars-replaced nil))
+         (is-after-change (not is-before-change))
+         ;; for a deletion, the post-change beginning and end are at the same place.
+         (is-insertion (and is-after-change (not (equal beg end))))
+         (is-deletion (and is-before-change (not (equal beg end)))))
+    (when (or is-insertion is-deletion)
+      (let* ((range-start (list :line (- (line-number-at-pos beg) copilot--line-bias)
+                                :character (- beg (save-excursion (goto-char beg) (line-beginning-position)))))
+             (range-end (if is-insertion range-start
                           (list :line (- (line-number-at-pos end) copilot--line-bias)
                                 :character (- end (save-excursion (goto-char end) (line-beginning-position))))))
-
+             (text (if is-insertion (buffer-substring-no-properties beg end) ""))
              (content-changes (vector (list :range (list :start range-start :end range-end)
                                             :text text))))
-
         (cl-incf copilot--doc-version)
         (copilot--notify 'textDocument/didChange
                          (list :textDocument (list :uri (copilot--get-uri) :version copilot--doc-version)
-							   :contentChanges content-changes))))))
+                               :contentChanges content-changes))))))
 
 (defun copilot--on-doc-close (&rest _args)
   "Notify that the document has been closed."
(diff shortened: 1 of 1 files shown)


CeleritasCelery:
This looks great! Thank you for implementing this. Your solution looks cleaner then what I was planning. I especially appreciate that we are no longer sending the entire buffer to copilot but only a small change set as needed. I tested it and it seems to work well.

CeleritasCelery:
I was doing some more testing on this, and I noticed some issues.

First is that `window-selection-change-functions` is called both when we switch to a copilot buffer *and* when we switch away from one, so we are calling `didFocus` for a whole bunch of non-copilot buffers (like the minibuffer or message buffer). These are also being added to `copilot--opened-buffers`, which they should not be.

Second is that disabling `copilot-mode` in a buffer is not removing that buffer from `copilot--opened-buffers`, so in that list when it should not be.

vdwemil95:
@CeleritasCelery You are correct on both counts. Looks like it definitely needs a bit more work which I'm hoping to get done this week. It should not be that difficult to remedy though.

Thanks for your diligent testing and great feedback. Really appreciate it! :trophy:

emil-vdw:
I have to do some more testing but it seems that the mode cleanup body is called for every open buffer when leaving the global mode so it should be fine.

Should we send `didClose` event when deactivating `copilot-mode`? Is there any other cleanup we need to do when leaving the mode?

CeleritasCelery:
> Should we send didClose event when deactivating copilot-mode?

I would say yes. We call `didOpen` when activitating `copilot-mode`, so it makes sense to be symmetrical. From copilots perspective, that buffer is gone.

One other thing I noticed in testing this change is that `window-selection-change-functions` is not called when we change the buffer within a window (for example, using `previous-buffer`, `next-buffer`). This means we don't get a `didFocus` event.

emil-vdw:
I agree. I will implement the `didClose` signal on mode exit.

As for the `previous-buffer`, `next-buffer` case, it looks like hooking onto both `window-selection-change-functions` and `window-buffer-change-functions` seems to cover all bases. It does not look like they overlap but I will double check.

Also looks like `window-selection-change-functions` hooks are called twice. Once with the window being switched away from as the argument and again with the window switched to as argument. The active buffer is the new (switched to) buffer in either case so no bug there; however, it does mean that we sometimes send `didFocus` events twice. I'll see if I can ignore the first event of the two in that case.

CeleritasCelery:
I found another issue as I was doing some testing. With the new change, the only time we send the entire buffer to copilot is on the call to initial call to `didOpen` using `copilot--get-source`. However `copilot--get-source` has default limit of 30000 characters. meaning if your buffer is over that size, copilot never receives part of the buffer.

I noticed this issue because sometimes copilot would just not be sending me any completions, but if I called the old function `copilot--sync-doc` it would start working again. The issue was that I was working outside of this 30000 character boundary, and so copilot had no idea what the context was, because we never gave it to it.

emil-vdw:
> However copilot--get-source has default limit of 30000 characters

Could we not significantly increase the default limit? 
It's not being called on every keypress anymore right.

CeleritasCelery:
We could, and I have set it to -1 in my case so that it has no limit. But I am still observing the same behavior. copilot will stop sending me completions until I call `copilot--sync-doc`. I am not sure if this is something only effecting me or a problem with how I have copilot configured.

emil-vdw:
> copilot will stop sending me completions until I call `copilot--sync-doc`.

It's not just you. I've been having the exact same problem. Will debug this and report back as well.

CeleritasCelery:
I wonder if this is just a case of copilots internal buffer getting screwed up when we accept completions. For example when we accept a completion we send `notifyAccepted`, but we also send the just accepted text via `textDocument/didChange`. If copilot adds the text to it's internal buffer when we call `notifyAccepted` (I don't know that it does this, I am just speculating) then it is inserting two copies of the text at the same place. This would really screw things up.

Also keep in mind that we post-process the completion text via `copilot--show-completion` (by both removing shared prefixes and balancing parens). If copilot is updating its internal model on `notifyAccepted`, this would cause problems as well.

emil-vdw:
What you're saying makes a ton of sense... I'm going to use copilot today with `notifyAccepted` disabled to see if that makes a difference.

At this point I'm just going to try to inspect the agent's internal state... Going to start here: [https://github.com/haukot/copilot_specifications#how-to-debug](url).

emil-vdw:
The node debugger route wasn't very fruitful. Still testing things out with `notifyAccepted`, `notifyRejected` disabled.

Another potential bug I've uncovered is that event `uri` for any buffer that is not saved to file is an empty string. This will garble the shared empty string URI buffer for copilot buffers not saved to file yet (\*scratch\* buffer for example)

emil-vdw:
Another bug I've found is that in `copilot--get-uri` the buffer local variable `buffer-file-name` is used to determine the `uri`. However, there is a brief moment of time when this has not been set yet (I have `copilot-mode` hooked to `prog-mode`) and defaults to `uri` of empty string. 

If a change happens in this time window the version increases with `uri` empty string until the var is set and the 'file://...' `uri` is used. This also garbles the internal state / creates a version mismatch which causes prediction requests to error.

MtCoin406 (approved):

emil-vdw:
@CeleritasCelery @MtCoin406 I think this is ready for final review and merge. 

I will log a separate issue for change in URI when saving an active copilot buffer that was not saved to file before.

zerolfx:
There might be a bug. The content of company documentation is being sent as a change to the Python file.
image
image

zerolfx:
Will there be any issues if I enable `copilot-mode` while editing the current buffer? The buffer may not be opened in Copilot.

emil-vdw:
> Will there be any issues if I enable `copilot-mode` while editing the current buffer? The buffer may not be opened in Copilot.

No issues. When activating the mode, the on focus handler is triggered manually which will send the `didOpen` notification.

emil-vdw:
> There might be a bug. The content of company documentation is being sent as a change to the Python file. image image

I will look into this and see if I can fix it.

emil-vdw:
@zerolfx I'm not able to reproduce this, could you give me a recipe config maybe? I'm using company and company box but even without company box I can't get anything to modify the actual buffer nor send a copilot event.

MtCoin406:
Access

zerolfx:
@emil-vdw I will review and make any possible fixes this weekend. Hopefully, we can get this PR merged then.
I greatly appreciate your work on this.

emil-vdw:
Thanks @zerolfx. That would be great. Let me know if there's anything else I can do. Would love to make more contributions in future.

CeleritasCelery:
@emil-vdw what was the fix for the "copilot not sending completions until re-synced" issue?

emil-vdw:
@CeleritasCelery there were change events that were replacements and not just an insertion or deletion. One example of this is when yanking from the kill ring, I think it inserted the text and then replaced it with text that has the correct text properties. 

Just had to modify the before and after change hooks to process a deletion for the replaced text in the before hook, and an insertion for the replaced text in the after hook.

zerolfx:
@emil-vdw 
I give you write access to this repo. I greatly appreciate your help with this, as I no longer use Emacs in my daily work.

CeleritasCelery:
I have been using the master branch with these changes in it, but I am still experiencing the issue where copilot looses stare until I sync it explicitly.

PR #167, opened 2023-08-12
Modify copilot--base-dir to accomodate straight.el/elpaca

ultronozm:
I encountered errors when attempting to use copilot.el with either straight.el or elpaca.  The problem turned out to be that copilot--base-dir was computed to be the "build" directory containing the symlinked source files, rather than the "repo" directory containing the other assets.  The fix is just to follow the symlink before computing the directory.

### DIFF
--- a/copilot.el
+++ b/copilot.el
@@ -56,47 +56,22 @@ Enabling event logging may slightly affect performance."
   :group 'copilot
   :type 'integer)
 
-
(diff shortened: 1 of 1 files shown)


zerolfx:
What does your straight configuration look like?
Did you forget to specify `:files ("*.el" "dist")`?

ultronozm:
Thanks!  Yes, that was the issue.

Issue #172, opened 2023-08-29
PR 164 breaking org-todo


shouya:
Today I noticed my org-mode started to behave abnormally. The observed behavior is that when marking a heading as DONE with `org-todo` (`C-c C-t`), the heading did not get correctly marked as expected.

```
# before running `org-todo` (point on the heading)
* TODO heading

# after running `org-todo`
*heading

# expected state
* DONE heading
```

It also errors with `args-out-of-range` message at [this position](https://github.com/bzg/org-mode/blob/main/lisp/org.el#L9383).

I wasn't suspecting `copilot.el` initially, but bisecting my config points to `copilot.el` as the culprit. This is double confirmed by disabling `copilot-mode` on the org buffer, and then the problem is gone.

I then scanned `copilot.el` commits and find that it's the most recent commit (#164) that's causing the problem. Before this version everything is fine. I have confirmed the problem on two org versions, namely `9.7-pre` (latest) and `9.6.6` (emacs 29.1 built-in).

I'm suspecting the added `{before,after}-change-functions` hooks may be conflicting with `org-todo`. That said, I'm expecting the change may break other places as well but I didn't observe any (yet). Sorry that I don't possess the sufficient skill to spot out how exactly how it's broken and patch it.

shouya:
cc @emil-vdw.

Thanks for that great PR btw, I was really looking forward for this feature - that's why I updated copilot.el as soon as it was merged :)

emil-vdw:
@shouya Great to hear! Will look into it this weekend.

In the meantime, would you please enable event logging by customize copilot-log-max (to e.g. 1000), then paste related logs in the *copilot events* and *copilot stderr* buffer.

shouya:
Hi @emil-vdw, here's the `*copilot events*` buffer you asked for.

```
[client-notification] Thu Aug 31 00:44:17 2023:
(:jsonrpc "2.0" :method "textDocument/didFocus" :params
          (:textDocument
           (:uri "file:///home/shou/tmp/scratch/scratch-1LZkZB.org")))
[client-notification] Thu Aug 31 00:44:19 2023:
(:jsonrpc "2.0" :method "textDocument/didFocus" :params
          (:textDocument
           (:uri "file:///home/shou/tmp/scratch/scratch-1LZkZB.org")))
[client-notification] Thu Aug 31 00:44:21 2023:
(:jsonrpc "2.0" :method "textDocument/didChange" :params
          (:textDocument
[...]

PR #176, opened 2023-09-01
[Improvement] Defer sending doc changes to the agent to an idle timer.

raymond-w-ko:
Currently, doc changes (like `textDocument/didChange`) are communicated on every change (essentially every key press). It would make more sense and improve responsiveness to defer communicating with the server until some idle time or until it is actually needed for a completion.

To implement this we don't have to reinvent the wheel and can just import most of it from the `eglot` implementation. 

## Implementation
- Use a change queue to store doc changes.
- Flush doc changes on idle timer.
- Flush doc changes before making a completion request.
- Don't send any pending events after the document has been closed or copilot mode has been disabled for a that buffer.

This fixes issue:
https://github.com/zerolfx/copilot.el/issues/172 (although this is also seems to not be an issue in newer `org` versions).


raymond-w-ko:
The original pull request would never have worked properly, since I thought `nreverse` reversed in place, but it was actually dropping the subsequent set of changes, meaning the buffer would have been desynced.

There is also some other extra stuff here, like the `(copilot-resync-buffer)` command, and a rewrite to fix `save-*` nesting order.

But assuming this is the wrong way to fix #172, there isn't anything too important here to merge.

emil-vdw:
> But assuming this is the wrong way to fix #172, there isn't anything too important here to merge.

@raymond-w-ko I am confused. does this PR fix the linked issue or not?

emil-vdw (changes requested):

emil-vdw:
I will push changes and get this merged.

copilot.el
```diff
@@ -759,20 +776,44 @@ Use TRANSFORM-FN to transform completion if provided."
...
+            (cl-incf copilot--doc-version)
+            (push (list :textDocument (list :uri (copilot--get-uri) :version copilot--doc-version)
+                        :contentChanges content-changes)
+                  copilot--doc-change-queue)))))
+
+    ;; If this is the after change hook, we need to set/reset the idle timer
+    ;; that will flush the pending changes to the copilot agent.
+    (when is-after-change
+      ;; If the idle timer is already set, cancel it and set a new one.
+      (when copilot--change-idle-timer (cancel-timer copilot--change-idle-timer))
```
jcs090218, line 804: You might to use the function `timerp` here.

copilot.el
```diff
@@ -759,20 +776,44 @@ Use TRANSFORM-FN to transform completion if provided."
...
+              (run-with-idle-timer
+               copilot-send-changes-idle-time
+               nil (lambda ()
+                     ;; Only flush the pending changes if the buffer is still
+                     ;; alive and in copilot mode.
+                     (when (buffer-live-p buf)
+                       (with-current-buffer buf
+                         (when copilot-mode
+                           (copilot--flush-pending-doc-changes)
+                           (setq copilot--change-idle-timer nil)))))))))))
```
jcs090218, line 816: Why are we setting the timer to `nil` here? 🤔 I prefer not to nullify the timer, even if that's the end of execution.

copilot.el
```diff
@@ -751,6 +755,19 @@ Use TRANSFORM-FN to transform completion if provided."
...
+Each element is a list of document change parameters to be sent to the agent.")
+
+(defvar-local copilot--change-idle-timer nil
+  "Idle timer for didChange signals.")
+
+(defun copilot--flush-pending-doc-changes ()
+  "Flush the pending document changes to the copilot agent."
+  (dolist (doc-change (nreverse copilot--doc-change-queue))
+    (copilot--notify 'textDocument/didChange doc-change))
+  (setq copilot--doc-change-queue '()))
```
jcs090218, line 769: It's better just to set it to `nil`.

copilot.el
```diff
@@ -117,6 +117,10 @@ You may adjust this variable at your own risk."
   :type 'string
   :group 'copilot)
 
+(defcustom copilot-send-changes-idle-time 0.5
+  "Don't tell server of changes before Emacs's been idle for this many seconds."
+  :type 'number)
```
jcs090218, line 122: Missing the property `:group`.

copilot.el
emil-vdw: Then the `dolist` call will error. Could wrap it in a check though.

copilot.el
emil-vdw: Honestly, I took this code straight from the `eglot` implementation. Probably good to clean up and have a easy way to know if the timer has finished execution?

copilot.el
jcs090218: What error is it? It shouldn't give you any error. 🤔 All these safely executed:

```elisp
(dolist (_ (nreverse nil)))
(nreverse nil)
(dolist (_ nil))
```

emil-vdw: Fair enough! :smile:

copilot.el
jcs090218: See the function description from `cancel-timer`:

```
Remove TIMER from the list of active timers.
```

Setting the variable to `nil` does not seem to remove the timer. I assumed the timer list would grow indefinitely.

emil-vdw: The timer doesn't need to be cancelled because it has already executed and it is non-repeating (the var containing timer is set to nil in the timer function body).

jcs090218: You are right. The timer will be removed by itself. 🤔 Thanks for the confirmation!

jcs090218 (approved):
LGTM. :)

Some nits that are good to resolve:

```
In toplevel form:
copilot.el:39:35: Warning: Alias for ‘copilot-completion-idle-delay’ should be declared before its referent

In copilot-diagnose:
copilot.el:311:7: Warning: reference to free variable ‘copilot-mode’

In copilot--handle-notification:
copilot.el:542:14: Warning: ‘mark-whole-buffer’ is for interactive use only.

In copilot--on-doc-change:
copilot.el:665:32: Warning: reference to free variable ‘copilot-mode’

In copilot--on-doc-focus:
copilot.el:675:14: Warning: reference to free variable ‘copilot-mode’

In copilot-async-start-process:
copilot.el:[105](https://github.com/copilot-emacs/copilot.el/actions/runs/8222298553/job/22483626619?pr=176#step:5:106)0:10: Warning: Unused lexical variable ‘name’

In end of data:
copilot.el:804:14: Warning: the function ‘vterm-insert’ is not known to be defined.
copilot.el:803:14: Warning: the function ‘vterm-delete-region’ is not known to be defined.
copilot.el:543:14: Warning: the function ‘org-sort-entries’ is not known to be defined.
copilot.el:532:54: Warning: the function ‘org-entry-get’ is not known to be defined.
copilot.el:532:26: Warning: the function ‘org-map-entries’ is not known to be defined.
```

But we can fix these warnings later.

emil-vdw:
One last thing to check is that `copilot-panel-complete` also flushes pending messages before requesting completions.

copilot.el
```diff
@@ -26,7 +26,7 @@
   :group 'completion
   :prefix "copilot-")
 
-(defcustom copilot-idle-delay 0
+(defcustom copilot-completion-idle-delay 0
```
timcharper, line 29: what's the motivation for this change?

copilot.el
```diff
@@ -565,6 +575,121 @@ automatically, browse to %s." user-code verification-uri))
     (org-mode)
     (erase-buffer)))
 
+;;
+;; Doc changes
+;;
+(cl-defmacro copilot--widening (&rest body)
```
timcharper, line 581: It feels like this change doens't quite fit the PR description? Separate PR? Not clear to me why it's necessary since the point should be within the narrowed region, right?

timcharper (changes requested):
Change appears to not create any regressions, however, I'm unable to reproduce the original issue on master, which causes me to wonder if we really understand what is broken?

[asciicast](https://asciinema.org/a/rehfVyeT0q7An9g6voAUkperv)

I have concerns about the widen change. Why is this introduced with this PR? There's no explanation as to what motivated it, and it appears unrelated and also I cannot understand why it is needed (copilot operates right after the point, and the point should always be in a narrowed region? In my experiments, I have no trouble using copilot within a narrowed region. A separate issue should be filed.

[asciicast](https://asciinema.org/a/pVuFhZuBsxtIromX1xvl49fTo)

In my recordings, I'm using Emacs 29.2, with the built-in version of org-mode, and a minimal environment (package-initialize, keycast-mode enabled, copilot-mode enabled (from source at commit 0d48e75dabef86b962b7a8bfcc7d3614a1201821)

Given that I can't reproduce the original issue on master, I'm wondering if this change is needed

emil-vdw:
@timcharper Although I did manage to reproduce the problem initially I am also not seeing it anymore on Emacs `29.2` org mode version `9.6.15`.

I do, however, feel like this is a good improvement nonetheless. That being deferring jsonrpc communication with the server to some idle time or until the user requests a completion (as implemented in and shamelessly stolen from `eglot`). This can still be "turned off" by setting the idle time low enough.

In any event, the PR title and description needs some work. I will update.

bbatsov:
I guess this should account for the changes in #363.

bbatsov:
@raymond-w-ko @emil-vdw Is this something you plan to continue working on or should we close it?

emil-vdw:
@bbatsov this needs a bit of work still and I'm not sure when I'll get back around to it. I've also been considering re-implementing this with the newer and simpler `track-changes` API.

I think you can close this for the time being. I will create a new PR in future :+1:.

PR #183, opened 2023-09-13
Use cl-oddp from cl-lib

byronclark:
oddp is undefined in at least Emacs 29.1.

This prevents a storm of messages like these while editing strings and especially comment strings in Clojure.
```
error in process filter: Symbol’s function definition is void: oddp
error in process filter: copilot-balancer-odd-dquote-count-p: Symbol’s function definition is void: oddp
```


raymond-w-ko:
Ahh, sorry about that. I just began learning about Emacs like two months ago. 😔

Issue #179, opened 2023-09-02
indentation issue with overlays and completions


ultronozm:
Put the following in an elisp buffer:

```
(defun print-sum (a b)
  (let ((sum (+ a b)))
```

Position the cursor at the end of the second line and do `M-x copilot-complete`.  An overlay with `(print sum)` appears on the next next with the correct indentation.  If I accept the completion, I get:
```
(defun print-sum (a b)
  (let ((sum (+ a b)))
    (print sum)))
```

So far, so good.  Now, go back to the initial buffer contents and insert a newline.  The buffer should then look as follows, with "|" indicating cursor position:

```
(defun print-sum (a b)
  (let ((sum (+ a b)))
    |
```
Now do `M-x copilot-complete` (if necessary).  Observe that the suggested completion has its overlay indented too much (the same amount as the current column).  If we do `M-x copilot-accept-completion`, then we get:
``` elisp
(defun print-sum (a b)
  (let ((sum (+ a b)))
        (print sum)))
```

It's easy to patch things up afterwards via `M-q`, etc., but that seems suboptimal.  I suspect that the copilot module "knows" the correct indentation, but that `copilot.el` somehow fails to "subtract" the indentation of point?  Haven't tried to debug this yet.

raymond-w-ko:
Does merged PR #178 fix this?

ultronozm:
> Does merged PR #178 fix this?

No.  I'm on the latest version, and the issue is anyway unrelated to narrowing.

raymond-w-ko:
Sorry, I got confused with issue #173 for some reason. 😞

ultronozm:
I think the following example rules out some other possibilities:

``` elisp
(defun print-hello-then-print-sum (a b)
  (let ((sum (+ a b)))
    (print "hello")
```

Position the cursor at the beginning of the next line, all the way to the left, like where "|" is in the following:
``` elisp
(defun print-hello-then-print-sum (a b)
  (let ((sum (+ a b)))
    (print "hello")
|
```

Then do `M-x copilot-complete`.  Get similar issue as above:
``` elisp
(defun print-hello-then-print-sum (a b)
  (let ((sum (+ a b)))
    (print "hello")
        (print sum)))
```

Using edebug on `copilot--show-completion` and inspecting the argument `completion-data` shows that the error has already occurred:

``` elisp
(:uuid "27fc2ffc-6e2d-41c2-9b23-e116eb3008bf"
       :text "        (print sum)))"
[...]

PR #191, opened 2023-10-10
print warning message when copilot finds no mode-specific offset

ultronozm:
See https://github.com/zerolfx/copilot.el/issues/179.  This is my best attempt to implement the suggestion made in the comment https://github.com/zerolfx/copilot.el/issues/179#issuecomment-1710276275.


blahgeek:
Similar to https://github.com/zerolfx/copilot.el/pull/198, I think we should not invent our own "warning suppression" method.

ultronozm:
I agree with your suggestion, and wasn't aware of the standard way to
communicate warnings.  Thanks!  (As far as I'm concerned, please feel free
to PR this if you know how to do it the right way)

blahgeek:
Though I'm not sure if it's ok to simply remove the `copilot-indent-warning-suppress` option since it's a "public" option that is already in master? Or do we need some deprecation process?

ultronozm:
Given that I both reported the issue and submitted the fix, I would guess that not too many people are using this setting, so you shouldn't worry too much about having to properly deprecate it.

Issue #205, opened 2023-12-01
The bot is not catching up copilot.vim releases anymore


rakotomandimby:
[Copilot.vim has run away](https://github.com/github/copilot.vim/tags) and [the last release caught by the bot is 1.10.2](https://github.com/zerolfx/copilot.el/commit/5a613b5d074b2811c95637811595b40283e21b1a).
There is a problem with the bot.

rakotomandimby:
@zerolfx , 

1. I tested on my daily use case, not extensive tests.
2. I will be honoured to try to continue your work.

PR #207, opened 2023-12-03
Catching  copilot.vim v1.12.1 up

rakotomandimby:
Attempt to resolve https://github.com/zerolfx/copilot.el/issues/205


csheaff:
submodule?

csheaff:
@rakotomandimby thanks for the response. I suppose I'll just compare the hashes of the binaries with the official ones.

zerolfx:
@csheaff Thanks for your suggestion.
The primary reason I don't use git submodule is to avoid complicating the package installation process, especially with those different package managers.

PR #212, opened 2023-12-17
Use `display-warning` to print indent warnings

blahgeek:
See comments in https://github.com/zerolfx/copilot.el/pull/191

### DIFF
--- a/copilot.el
+++ b/copilot.el
@@ -291,14 +291,6 @@ Enabling event logging may slightly affect performance."
 (defvar-local copilot--completion-cache nil)
 (defvar-local copilot--completion-idx 0)
 
-(defcustom copilot-indent-warning-suppress nil
-  "If nil, then warn when copilot finds no mode-specific offset."
-  :type 'boolean
-  :group 'copilot)
-
-(defvar-local copilot--indent-warning-printed-p nil
-  "Flag indicating whether indent warning was already printed.")
-
 (defun copilot--infer-indentation-offset ()
   "Infer indentation offset."
   (or (let ((mode major-mode))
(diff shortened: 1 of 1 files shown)


fgiasson:
What was the reason to remove `copilot-indent-warning-suppress`? 

In my current setup, I am getting warning messages every time that I type something while being in `copilot-mode`. Is there a new way to suppress them? (can't see from this PR)

Ramblurr:
What is the proper fix for this warning?

Ramblurr:
image

This new warning that is displayed upon every keypress is rather frustrating.

1. It is displayed on every keypress
2. It does not contain (nor the docs here on github) any actionable instructions

This renders copilot.el effectively useless and the warning is basically a fatal error because I can't use emacs with it enabled.

If its just a warning (i.e., a non fatal error that can be ignored) it shouldn't display on every keystroke.

And ideally there would be some actionable advice or documentation about how to resolve the warning/error.

blahgeek:
@Ramblurr sorry for the inconvenience. Printing the warning on each keystroke is not expected, I will fix it ASAP. To suppress the warning, execute `(add-to-list 'warning-suppress-types '(copilot copilot-no-mode-indent))` (or to suppress all warnings from copilot, `(add-to-list 'warning-suppress-types 'copilot)`)

Ramblurr:
Thanks @blahgeek I appreciate the response.

Aside from suppressing the warning is there a proper way to fix the problem (which I don't fully understand tbh)?

blahgeek:
Well, according to the code, it warns because current major mode's indentation cannot be found in neither `copilot--indentation-alist` or `editorconfig-indentation-alist`. Adding corresponding entries to them should be able to fix it.

More details in https://github.com/zerolfx/copilot.el/issues/179

PR #222, opened 2023-12-27
Fix jumping cursor bug

### LINKED ISSUE #200: Cursor position error
When the real cursor position is not at the start of the overlay preview. The cursor will jump to the end of the overlay.

For example:
copilot_bug
the cursor will keep jumping to the end of the overlay.

Another similar situation:
another_bug
the cursor's real position is after :, but it will jump to the end of the completion.

To reproduce:

```shell
emacs -Q -l test.el test.py
```

```elisp 
;; test.el
(package-initialize)

;; straight.el initialization
(defvar bootstrap-version)
(let ((bootstrap-file
       (expand-file-name "straight/repos/straight.el/bootstrap.el" user-emacs-directory))
      (bootstrap-version 6))
  (unless (file-exists-p bootstrap-file)
    (with-current-buffer
        (url-retrieve-synchronously
         "https://raw.githubusercontent.com/radian-software/straight.el/develop/install.el"
         'silent 'inhibit-cookies)
      (goto-char (point-max))
      (eval-print-last-sexp)))
  (load bootstrap-file nil 'nomessage))

(add-to-list 'package-archives
             '("melpa" . "https://melpa.org/packages/"))

;; Location of my use-package git clone
(add-to-list 'load-path "~/.emacs.d/.local/straight/repos/use-package/")
(require 'use-package)

(use-package copilot
  :straight (:host github :repo "zerolfx/copilot.el" :files ("dist" "*.el"))
  :ensure t)

(add-hook 'prog-mode-hook 'copilot-mode)
```

emil-vdw:
Fix a bug where the cursor would jump to the end of the overlay when the completion candidate starts with a newline and the cursor is at the end of the current line.

This happens because setting the `'cursor` property on a newline does not make the cursor show at the end of the line.

Implementation:
- Refactor `copilot--set-overlay-text` function.
- Add a "ghost" whitespace character to focus the cursor on in this edge case.

Closes #200.

### DIFF
--- a/copilot.el
+++ b/copilot.el
@@ -525,19 +525,27 @@ To work around posn problems with after-string property.")
 
 (defun copilot--set-overlay-text (ov completion)
   "Set overlay OV with COMPLETION."
+  ;; Clear old overlay content.
+  (overlay-put ov 'display "")
+  (overlay-put ov 'after-string "")
+  (setq copilot--real-posn (cons (point) (posn-at-point)))
   (move-overlay ov (point) (line-end-position))
+
   (let* ((tail (buffer-substring (copilot--overlay-end ov) (line-end-position)))
-         (p-completion (concat (propertize completion 'face 'copilot-overlay-face)
-                               tail)))
-    (if (eolp)
-        (progn
-          (overlay-put ov 'after-string "") ; make sure posn is correct
-          (setq copilot--real-posn (cons (point) (posn-at-point)))
-          (put-text-property 0 1 'cursor t p-completion)
-          (overlay-put ov 'display "")
-          (overlay-put ov 'after-string p-completion))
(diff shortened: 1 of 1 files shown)


zerolfx:
You can merge it without approval or directly commit to the main branch.

zerolfx:
The Emacs overlay is so tricky, and I'm unsure if a perfect solution exists.

david-krentzlin:
This commit cause my setup to break. Working within double quotes will cause the cursor to jump behind the closing quote, which makes working with copilot impossible. The same is true for braces when in lisp mode. It could be a combination of other things in my setup, which is why I just left a comment and didn't open a bug ticket. 

I've pinned my package to the previous commit for now, which fixes the issue for me.

PR #224, opened 2023-12-29
fix: Deprecated events-buffer-scrollback-size

AlexanderArvidsson:
Deprecation was introduced in `jsonrpc 1.0.22` in [this commit](https://github.com/emacs-mirror/emacs/commit/e0b9944b69ff72923c29756fcfcea9528a3f5069).

This PR ~bumps required version to `1.0.22` and~ now uses recommended `:events-buffer-config`.
It's backwards compatible by checking for existence of `-events-buffer-config` slot on `jsonrpc-process-connection`.

### DIFF
--- a/copilot.el
+++ b/copilot.el
@@ -1,6 +1,6 @@
 ;;; copilot.el --- An unofficial Copilot plugin for Emacs  -*- lexical-binding:t -*-
 
-;; Package-Requires: ((emacs "27.2") (s "1.12.0") (dash "2.19.1") (editorconfig "0.8.2") (jsonrpc "1.0.14"))
+;; Package-Requires: ((emacs "27.2") (s "1.12.0") (dash "2.19.1") (editorconfig "0.8.2") (jsonrpc "1.0.23"))
 
 ;;; Code:
 (require 'cl-lib)
@@ -139,20 +139,6 @@ Enabling event logging may slightly affect performance."
                                               (funcall ,success-fn result)))
                               ,@args))))
 
-(defun copilot--slot-exists-p (slot-list slot-name)
-  "Return t if SLOT-NAME exists in SLOT-LIST, nil otherwise."
-  (cl-some (lambda (slot) (eq slot-name (cl--slot-descriptor-name slot))) slot-list))
-
-(defun copilot--make-process ()
-  "Make a process for copilot agent."
-  (make-process :name "copilot agent"
-                :command (list copilot-node-executable
-                               (concat copilot--base-dir "/dist/agent.js"))
-                :coding 'utf-8-emacs-unix
-                :connection-type 'pipe
-                :stderr (get-buffer-create "*copilot stderr*")
-                :noquery t))
-
 (defun copilot--start-agent ()
   "Start the copilot agent process in local."
   (if (not (locate-file copilot-node-executable exec-path))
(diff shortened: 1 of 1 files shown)


AlexanderArvidsson:
Unsure if the change was introduced in `1.0.23` or in `1.0.22`. The change is there in `1.0.22` on git, but that doesn't mean that release had the change in it.

To be safe, perhaps set required `1.0.23` instead?

ethan-leba:
We should probably maintain backwards compat since `jsonrpc` is built-in?

AlexanderArvidsson:
@ethan-leba I actually tried that, but it ended up becoming super difficult.
My first thought was to detect if `:events-buffer-config` was available, but this was not possible without doing some pretty advanced `eieio` stuff.
My second thought is to try and detect which version of `jsonrpc` is installed, and conditionally change it.

However, conditionally choosing the keyword to use is really beyond my elisp knowledge.
We could duplicate the entire `make-instance` call, and conditionally do it based on some condition (either with the complex `eieio` or `jsonrpc` version check).

If I had to do backwards compatibility, I would try to get the `jsonrpc` version, then duplicate the `make-instance` call to do different things. I'm willing to attempt this, unless you have a better idea how to conditionally apply different keywords based on some condition.

AlexanderArvidsson:
So it is possible to get the `jsonrpc` version via `(package-version-join (package-desc-version (car-safe (cdr (assq 'jsonrpc package-archive-contents)))))`, however, this will error if `package--initialized` is nil, and initializing `package.el` takes quite a bit of time.

As such, I'm not comfortable relying on `package.el` to give me the `jsonrpc` version because it would slow down opening the first file after starting emacs. Either there's a different way, or a different condition needs to be chosen.

AlexanderArvidsson:
@ethan-leba Pushed changes for backwards compatibility based on inspecting the slots via `eieio-class-slots`.
Please take a look. This stuff is kind of new to me, so if you have a better way of checking for existence for `events-buffer-config` I'm happy to change it!

I haven't verified that this works on emacs installations that don't have `jsonrpc>=1.0.23` since I don't have time right now to create such an environment. But, I did check if my `copilot--slot-exists-p` method returns `t` and `nil` for `-events-buffer-config` and a non-slot respectively, so I'm pretty confident it should work.

When it comes to `eieio` dependency, do we need to add that to the `Package-Requires` section?

ethan-leba:
I think I was mistaken actually -- jsonrpc appears to be an ELPA package, not builtin so I think this current solution should work? https://elpa.gnu.org/packages/jsonrpc.html

Anyways, here's a good way to get the slots on a class:
```
(thread-last
  (find-class 'jsonrpc-process-connection)
  (eieio-class-slots)
  (mapcar (lambda (s) (cl--slot-descriptor-name s))))
```

AlexanderArvidsson:
@ethan-leba Actually, `jsonrpc` _is_ builtin. Check `list-package`, it shows entries for both ELPA and building:
```
  jsonrpc                        1.0.23         available    elpa     JSON-RPC library
  jsonrpc                        1.0.23.0.20231227.75345 available    gnu-devel JSON-RPC library
  jsonrpc                        1.0.16         built-in              JSON-RPC library
```

ethan-leba:
Oh interesting - regardless, I think the original solution should still be fine, the main concern would be if jsonrpc was tied to the version of emacs (which is about the only type of backwards compatibility anyone cares about in Emacs, really..)

AlexanderArvidsson:
So, do you prefer to revert the backwards compatibility I added (I could change the slot checking to your code, since it's much cleaner), or keep the backwards compatibility?

ethan-leba:
IMO, I think https://github.com/zerolfx/copilot.el/pull/224/commits/d47c7c603c1dd24d28a33a933df93764f3d458d5 would be best in the name of keeping code simple -- I'll defer to @zerolfx though.

AlexanderArvidsson:
Let's wait for @zerolfx's  response. In the meantime, I've simplified the slot checking to the following:
```lisp
(defun copilot--slot-exists-p (slot-list slot-name)
  "Return t if SLOT-NAME exists in SLOT-LIST, nil otherwise."
  (not (null (member slot-name (mapcar #'cl--slot-descriptor-name slot-list)))))
```

emil-vdw:
> IMO, I think [d47c7c6](https://github.com/zerolfx/copilot.el/commit/d47c7c603c1dd24d28a33a933df93764f3d458d5) would be best in the name of keeping code simple -- I'll defer to @zerolfx though.

I also do prefer the simplicity of making this (backwards incompatible) change. I don't see any major reason to maintain backwards compatibility.
Even for built in versions of `jsonrpc` < 1.0.23 a newer version is available from the gnu archive right?

In any event, `jsonrpc` version 1.0.23 provides backwards compatibility for the old keyword in this release even if a deprecation warning is displayed.

AlexanderArvidsson:
@emil-vdw @ethan-leba I reverted back to [d47c7c6](https://github.com/zerolfx/copilot.el/commit/d47c7c603c1dd24d28a33a933df93764f3d458d5), keeping the history. If you ever decide you want backwards compatibility, you can pick it up from [d00ba0d](https://github.com/AlexanderArvidsson/copilot.el/commit/d00ba0d5806c39af05edcdcee598a439f193b684).

Feel free to merge.

emil-vdw (approved):
Great work @AlexanderArvidsson, thanks for the fix!

xmatos-so:
Hello! Thank you so much for your work on this emacs package! The `jsonrpc` upgrade is incompatible with the stock `eglot` available with emacs 29.1. I understand the reasoning behind keeping the code simple, but I think you should use the standard `jsonrpc` package. What I'm doing right now is, I fetched `AlexanderArvidsson` remote repo, merged the latest from `main` and reverted **2395161**.

AlexanderArvidsson:
@xmatos-so Personally, I agree with you, and I don't think that the backward compatibility I had made the code too complex. Since `eglot` is incompatible, @emil-vdw perhaps it's better to go with the backward compatibility?

PR #245, opened 2024-01-28
test: Use Eask for CI

jcs090218:
I've added the basic test, see the result https://github.com/jcs-PR/copilot.el/actions/runs/7683725212.


emil-vdw (approved):
Very nice, good to have more contributors! :partying_face:

Issue #120, opened 2023-04-04
Add package to MELPA


jgarte:
Just opening this issue to track adding copilot.el to MELPA.

@zerolfx 

What do we have left to do?

zerolfx:
I am planning to use the [copilot-node-server](https://www.npmjs.com/package/copilot-node-server) NPM package (which is used by https://github.com/TerminalFi/LSP-copilot).

The primary tasks that must be completed are:
- Installing the NPM package on behalf of the user (in a similar manner to installing a language server in `lsp-mode`).
- Finding a method to check the current version and upgrade it if necessary.

Minor tasks include:
- Ensuring that current users can migrate to the new version
- Adding an open-source license
- Submitting the package to MELPA
- Adding CI for versioning, packaging, and other related tasks

zerolfx:
The potential problems are:
- `npm` becomes a new dependency
- users may not want to have an npm package installed globally (using `npm i -g` is the default behavior of `lsp-install-server`)
- it's difficult to check the version of a currently installed npm package
- more steps to setup `copilot.el`



There are a few potential issues to consider:

- `npm becomes a new dependency.
Some users may not want to install the npm package globally (the default behavior of lsp-install-server when installing language servers).
It can be challenging to check the current version of an installed NPM package.
Setting up copilot.el may require additional steps due to these changes.

rakotomandimby:
Node is already a dependency, which always embeds npm, AFAIK.

zerolfx:
> Node is already a dependency, which always embeds npm, AFAIK.

On Ubuntu and Arch, `npm` is not automatically installed when installing `nodejs`, as it's an optional dependency. But yes, we can assume `npm` is available if `node` is installed.

jgarte:
If I were to package this for Guix I would be patching [this line ](https://github.com/zerolfx/copilot.el/blob/main/copilot.el#L39) to something like the following:

`/gnu/store/gvwnkilh065b0sv9rr64xl50s0snjfl1-node-14.19.3/bin/node`

Here is some discussion about `copilot.el` in Guix:

https://issues.guix.gnu.org/62664#0

artlogic:
@zerolfx I noticed this issue was closed and the item is crossed out in the readme. Does this mean there are no plans to add this package to MELPA?

zerolfx:
@artlogic 
(The item was crossed out in the readme a long while ago.)
I have no plans to add the package to MELPA. However, I welcome any contributions.

[...]

PR #247, opened 2024-01-29
feat: Install server

jcs090218:
For #120.


copilot.el
```diff
@@ -841,10 +846,10 @@ Use this for custom bindings in `copilot-mode'.")
...
     (when (numberp copilot-idle-delay)
       (setq copilot--post-command-timer
-          (run-with-idle-timer copilot-idle-delay
-                               nil
-                               #'copilot--post-command-debounce
-                               (current-buffer))))))
+            (run-with-idle-timer copilot-idle-delay
+                                 nil
+                                 #'copilot--post-command-debounce
+                                 (current-buffer))))))
```
jcs090218, line 852: From L193 are all indentation errors. 🤔

copilot.el
```diff
@@ -97,8 +100,10 @@ indentation offset."
        (buffer-file-name)))
   "Directory containing this file.")
 
-(defconst copilot-version "0.10.0"
-  "Copilot version.")
+(defcustom copilot-version "1.14.0"
+  "Copilot version."
+  :type 'string
+  :group 'copilot)
```
jcs090218, line 106: I've made this variable customizable so the user can choose what version to install.

emil-vdw:
As part of this PR, we would also get rid of the `dist` folder and current `chore` workflow no?

zerolfx:
@emil-vdw 
We may need two steps:
1. Prioritize using the agent.js installed by the new method then the old dist folder. If a user is using the agent.js from the dist folder, send a deprecation warning and ask them to install it via the new method.
2. Remove the dist folder and the chore workflow.

zerolfx:
@jcs090218 
Could you also modify `copilot--start-agent` to load the `agent.js` from the new installing method? 
It would be better if you check the currently installed version of the NPM package and send a warning if it is mismatched with `copilot-version`.

zerolfx:
> Prioritize using the agent.js installed by the new method then the old dist folder. If a user is using the agent.js from the dist folder, send a deprecation warning and ask them to install it via the new method.

@emil-vdw After giving it another thought, I realized that this step can be skipped. It's okay to delete `dist` after providing a new way to install them.

jcs090218:
> We may need two steps:
> 
> 1. Prioritize using the agent.js installed by the new method then the old dist folder. If a user is using the agent.js from the dist folder, send a deprecation warning and ask them to install it via the new method.
> 2. Remove the dist folder and the chore workflow.

What if we prompt the user to install the server when it's not yet installed in the new location? This is much more straightforward IMO. 🤔 Then we can remove `dist` and `chore` workflow in this PR (or later).

> Could you also modify copilot--start-agent to load the agent.js from the new installing method? 
> It would be better if you check the currently installed version of the NPM package and send a warning if it is mismatched with copilot-version.

👍 Done.

zerolfx:
@jcs090218 Very cool! You solved the long-term issue!

One last request: 
Could you test the new installation method on Windows? (In case you have a Windows machine available.)

jcs090218:
> Could you test the new installation method on Windows? (In case you have a Windows machine available.)

Actually, I need someone to test on Unix-line system (linux or macOS), because I'm using Windows now. 😂

copilot.el
```diff
@@ -165,8 +176,15 @@ indentation offset."
...
-  (if (not (locate-file copilot-node-executable exec-path))
-      (user-error "Could not find node executable")
+  (cond
+   ((null copilot-node-executable)
+    (user-error "Could not find node executable"))
+   ((not (file-exists-p copilot-install-dir))
+    (user-error "Server is not installed, please install via `M-x copilot-install-server`"))
+   (t
+    (unless (equal (copilot-installed-version) copilot-version)
+      (message "[WARN] Newer versions of the Copilot server are available for installation.
```
emil-vdw, line 186: Why not warn here?
```suggestion
      (warn "Newer versions of the Copilot server are available for installation.
```

Also there are potentially some bigger issues with this approach. At some point we are going to add backwards incompatible changes (say chat functionality) that will just break if an older version of the server is installed?
For this same reason it makes it slightly risky for the user to be able to customize this var since it is somewhat coupled to our plugin implementation.

It might be worth considering whether we have to be more strict on this requirement (say by preventing copilot mode from activating unless some override var is set) or by automatically installing/upgrading the server?
@zerolfx @jcs090218

copilot.el
```diff
@@ -870,5 +888,73 @@ command that triggered `post-command-hook'."
...
+;; function for more information.
+(defun copilot-async-start-process (callback error-callback &rest command)
+  "Start async process COMMAND with CALLBACK and ERROR-CALLBACK."
+  (let ((name (cl-first command)))
+    (with-current-buffer (compilation-start (mapconcat #'shell-quote-argument (-filter (lambda (cmd)
+                                                                                         (not (null cmd)))
+                                                                                       command)
+                                                       " ") t
+                                                       (lambda (&rest _)
+                                                         (generate-new-buffer-name (format "*copilot-install: %s*" name))))
```
emil-vdw, line 915: We are not going to install a variety of different servers and the buffer name is confusing since we don't install npm.
```suggestion
                                                         (generate-new-buffer-name "*copilot-install-server*")))
```

copilot.el
```diff
@@ -182,7 +200,7 @@ indentation offset."
                                   :notification-dispatcher #'copilot--handle-notification
                                   :process (make-process :name "copilot agent"
                                                          :command (list copilot-node-executable
-                                                                        (concat copilot--base-dir "/dist/agent.js"))
+                                                                        (concat copilot--dist-dir "/agent.js"))
```
emil-vdw, line 203: On Linux, this doesn't work for me. There is an executable that is distributed by the npm package and changing it to this works for me:
```suggestion
                                                                        (f-join copilot-install-dir "bin" "copilot-node-server"))
```

emil-vdw (changes requested):
I have a few concerns:
1. The reliability of the `copilot-node-server` npm package. Can we trust it to be available, up-to-date and not to be hijacked? Should we investigate whether to pull the agent distributables straight from the corresponding github tag of the official `copilot.vim` repo?
2. The fact that at some point a mismatch between the minimum required server version and `copilot.el` version will break the plugin (see comment below).

copilot.el
jcs090218: I've changed the variable `copilot--dist-dir`. Does that work for your? 🤔 

```elisp
(defconst copilot--dist-dir
  (if (eq system-type 'windows-nt)
      (f-join copilot-install-dir "node_modules" "copilot-node-server" "copilot" "dist")
    (f-join copilot-install-dir "bin" "copilot-node-server"))
  "The dist directory containing agent.js file.")
```

copilot.el
jcs090218: I have a few thoughts on this:

1. Ignore it, but always support the latest version until the copilot is stable (less compatibility breakage).
2. Add another variable `copilot-preferred-version` and warn the user.
3. Document it somewhere on the README.md

I don't mind any approaches since there is no ideal solution, just personal preferences. 🤔

jcs090218:
> 1. The reliability of the copilot-node-server npm package. Can we trust it to be available, up-to-date and not to be hijacked? Should we investigate whether to pull the agent distributables straight from the corresponding github tag of the official copilot.vim repo?

I don't know much about the upstream `copilot.vim`, so I cannot suggest this. However, downloading npm packages seems very common in the Emacs world; use `lsp-mode` and `lsp-grammarly` as examples.

emil-vdw:
I'm not sure but those examples may be using official distributions (albeit through npm). The copilot vim is an official github owned (official copilot) repository so it can be relied upon. The npm package being installed by us through npm is not official and we should at least strongly consider pulling from the official distribution instead.

In my opinion there should be very good reasons for using the unofficial distribution instead.

jcs090218:
Ah, okay. I thought `copilot-node-server` is the official... How about publishing our own npm package? Like [lsp-grammarly](https://github.com/emacs-grammarly/lsp-grammarly) maintains its own [grammarly-language-server](https://github.com/emacs-grammarly/grammarly-language-server). 🤔

zerolfx:
I found the npm package `copilot-node-server` from https://github.com/TerminalFi/LSP-copilot a long while ago, which is another unofficial Copilot plugin for Sublime. The NPM package is maintained at https://github.com/jfcherng/copilot-node-server.

Maybe we can use it for now and later switch to our own NPM package.

jcs090218:
> Maybe we can use it for now and later switch to our own NPM package.

Yeah, I second this. It would take me some time to do this. Plus, it would make this PR too big if it was implemented in the same PR.

copilot.el
emil-vdw: It may be fine now the way it is if we document it well in the readme and the `defcustom` as well. That the default version should be modified at the users own risk and that the default on that defcustom is what is currently supported by `copilot.el`.

How does that sound?

jcs090218: I've documented in the `defcustom`, but I don't know where to document on the README. 🤔

emil-vdw: Actually, the `defcustom` comment is enough.

emil-vdw:
@jcs090218 I have added some commits to fix some bugs and support Linux. I don't have access to push to your branch though.

jcs090218:
Merged! :)

emil-vdw:
Would you mind testing those changes on Windows again as well?

jcs090218:
> Would you mind testing those changes on Windows again as well?

Yeah, it works now! :)

copilot.el
```diff
@@ -104,7 +104,8 @@ indentation offset."
 
 (defconst copilot--server-executable
   (if (eq system-type 'windows-nt)
-      (f-join copilot-install-dir "node_modules" "copilot-node-server" "copilot" "dist" "agent.js")
+      (f-join copilot-install-dir "node_modules" "copilot-node-server" "copilot"
+              "bin" "copilot-node-server")
```
jcs090218, line 108: I've simplified it using the same `bin/copilot-node-server` script.

copilot.el
```diff
@@ -245,27 +246,27 @@ Please upgrade the server via `M-x copilot-reinstall-server`"))
   "Login to Copilot."
   (interactive)
   (copilot--dbind
```
jcs090218, line 248: The rest are indentation errors.

copilot.el
```diff
@@ -179,8 +196,15 @@ indentation offset."
 
 (defun copilot--start-agent ()
   "Start the copilot agent process in local."
-  (if (not (locate-file copilot-node-executable exec-path))
-      (user-error "Could not find node executable")
+  (cond
+   ((null copilot-node-executable)
+    (user-error "Could not find node executable"))
+   ((not (file-exists-p copilot-install-dir))
```
emil-vdw, line 202: This did not catch the condition correctly when I uninstalled the copilot server (this directory is still present). Maybe make it more reliable?

readme.md
```diff
@@ -6,17 +6,17 @@ Copilot.el is an Emacs plugin for GitHub Copilot.
 
 **Warning:** This plugin is unofficial and based on binaries provided by [copilot.vim](https://github.com/github/copilot.vim).
 
-**Note:** You need access to [GitHub Copilot](https://github.com/features/copilot) to use this plugin.
+**Note:** You need access to [GitHub Copilot][] to use this plugin.
 
-Current maintainer: [@emil-vdw](https://github.com/emil-vdw), [@rakotomandimby](https://github.com/rakotomandimby).
+Current maintainer: [@emil-vdw][], [@rakotomandimby][].
```
emil-vdw, line 11:
```suggestion
Current maintainer: [@emil-vdw][], [@jcs090218][], [@rakotomandimby][].
```

readme.md
```diff
@@ -302,3 +302,13 @@ These projects helped me a lot:
 + https://github.com/TommyX12/company-tabnine/
 + https://github.com/cryptobadger/flight-attendant.el
 + https://github.com/github/copilot.vim
+
+
+<!-- Links -->
+
+[@emil-vdw]: https://github.com/emil-vdw
+[@rakotomandimby]: https://github.com/rakotomandimby
```
emil-vdw, line 310:
```suggestion
[@jcs090218]: https://github.com/jcs090218
[@rakotomandimby]: https://github.com/rakotomandimby
```

emil-vdw (approved):
Remember to pull main (jsonrpc error handling changes).

Thank you for all of your work on this project! You should have write access now and I will make sure you can merge without review (although try to get a review when practicable).

copilot.el
jcs090218: How did you uninstall the server? 🤔 The `copilot-install-dir` will be gone if you uninstall it through `M-x copilot-uninstall-server`.

emil-vdw: Ah that makes sense. I did:
```
npm -g --prefix ~/.emacs.d/.cache/copilot uninstall copilot-node-server
```

emil-vdw:
Shall we do the cleanup in a subsequent PR @jcs090218 @zerolfx?

emil-vdw:
I think it's fine to clean up in a subsequent PR. Feel free to merge :+1:.

sho-87:
this seems to install (or try to start) copilot server at the wrong location: 

```
[stderr] node:internal/modules/cjs/loader:1152
[stderr]   throw err;
[stderr]   ^
[stderr] 
[stderr] Error: Cannot find module 'c:\Users\simon\.emacs.d\.cache\copilot\node_modules\copilot-node-server\copilot\bin\copilot-node-server'
[stderr]     at Module._resolveFilename (node:internal/modules/cjs/loader:1149:15)
[stderr]     at Module._load (node:internal/modules/cjs/loader:990:27)
[stderr]     at Function.executeUserEntryPoint [as runMain] (node:internal/modules/run_main:142:12)
[stderr]     at node:internal/main/run_main_module:28:49 {
[stderr]   code: 'MODULE_NOT_FOUND',
[stderr]   requireStack: []
[stderr] }
[stderr] 
[stderr] Node.js v21.6.1
```

the extra `bin` before `copilot` leads to the wrong directory

jcs090218:
You are right. It's weird since I did applied the change but it some how got reverted. 🤔 

I'm on it.

jcs090218:
FYI, I've opened the fix in #267. However, I cannot merge PRs without other maintainer's approval. Therefore, the current workaround is to update `copilot--server-executable` manually. 

```elisp
(setq copilot--server-executable (f-join copilot-install-dir "node_modules" "copilot-node-server" "bin" "copilot-node-server"))
```

jmcarcell:
When opening emacs with this PR I get the following it complains that I should use `copilot-install-server` but when I try I get
`Symbol's value as variable is void: package`

jcs090218:
Can you try `toggle-debug-on-error`? And paste the backtrack here. Thanks!

jmcarcell:
```Debugger entered--Lisp error: (void-variable package)
  copilot-install-server()
  funcall-interactively(copilot-install-server)
  call-interactively(copilot-install-server record nil)
  command-execute(copilot-install-server record)
  execute-extended-command(nil "copilot-install-server" "install copilot")
  funcall-interactively(execute-extended-command nil "copilot-install-server" "install copilot")
  call-interactively(execute-extended-command nil nil)
  command-execute(execute-extended-command)
```
Indeed, I get this with main but don't get it if I checkout right before this PR

jcs090218:
I've made a fix in #268. Thanks for the bug report!

PR #248, opened 2024-01-29
Add jsonrpc dependency info

tninja:
Based on
1. https://github.com/copilot-emacs/copilot.el/issues/232
2. https://github.com/copilot-emacs/copilot.el/issues/244


tninja:
if there is a way to let copilot.el automatically install the dependency, it is much better than this.

jcs090218 (approved):

jcs090218:
What package manager are you using? The built-in package manager --- `package.el` will handle it for you. 🤔

tninja:
> What package manager are you using? The built-in package manager --- `package.el` will handle it for you. 🤔

Yes. Probably should replace ELPA to package manager.

tninja:
Move the description to installation part beside the editorconfig package dependency.

Issue #252, opened 2024-02-04
copilot-indentation-alist requires the indent value to be a symbol


kelvie:
e.g. you can't set ` (add-to-list 'copilot-indentation-alist '(org-mode 2)`, but if you use a defvar, like here: https://github.com/chloelee767/dotfiles/blob/feba957413eab9de115c8a46301208d708a76245/doom/.doom.d/config.el#L220-L227

It'll work.

jcs090218:
Is there any error when setting the variable? What's the value of `copilot-indentation-alist`? 🤔 I don't think `defcustom` and `defvar` would make any differences.

kelvie:
```
Debugger entered--Lisp error: (wrong-type-argument symbolp 2)
  #<subr F616e6f6e796d6f75732d6c616d626461_anonymous_lambda_35>(2)
  cl-some(#<subr F616e6f6e796d6f75732d6c616d626461_anonymous_lambda_35> (2))
  copilot--infer-indentation-offset()
  copilot--generate-doc()
  copilot--get-completion(#f(compiled-function (jsonrpc-lambda-elem1) #<bytecode 0xff2af3ebacb1718>))
  copilot-complete()
  copilot--post-command-debounce(#<buffer userscript.org>)
  apply(copilot--post-command-debounce #<buffer userscript.org>)
  timer-event-handler([t 0 0 0 nil copilot--post-command-debounce (#<buffer userscript.org>) idle 0 nil])
```

kelvie:
https://github.com/copilot-emacs/copilot.el/blob/2c1c425761e3cdaf139c9e2fa531f942cfb14b16/copilot.el#L325-L328

Looks like it's here, when the `alist-get` returns `2`, `cl-some` doesn't like that (never used common lisp before).

emil-vdw:
This should be fixed by #253 (thanks to @kelvie! :trophy:). I have tested this on my side but feel free to re-open if it is not resolved.

PR #253, opened 2024-02-04
Fix using numbers in copilot-indentation-alist

kelvie:
As mentioned in #252

Despite using emacs for almost 1.5 decades, my elisp still sucks so feel free to close this and fix it properly.

I've tested this still appears to work with both variable and number values in this alist.


emil-vdw:
Awesome! Thanks for fixing this. I will merge soon.

Issue #251, opened 2024-02-03
Sometimes, copilot.el's overlay is not considered an active overlay (and thus, the copilot-completion-map is not active)


timcharper:
Sometimes, Emacs does not consider copilot.el as an activate overlay (it never, for some reason, considers copilot.el's overlay in (overlays-at (point))

I use the elisp in this gist to help debug this:

https://gist.github.com/timcharper/fbb12d28bee0a75d8d6d54e463a5e324#file-copilot-debug-env-el

The elisp sets up copilot, keycast mode and a timer which updates a buffer *overlay-properties* every second with the following info:

* The current point position
* What the key TAB will do (because I have bound it to copilot-accept-completion in my copilot-completion-map)
* All the overlays in the current buffer, with their properties. Overlays considered "at the point" are highlighted.

I've tried my best to show the issue in the following asciinema:

[![asciicast](https://asciinema.org/a/43biwe6YjVmMEb5CCieNoNw6i.svg)](https://asciinema.org/a/43biwe6YjVmMEb5CCieNoNw6i)

# Screens

Screenshot showing that copilot-completion-map IS active if another overlay contains the point, but copilot.el's overlay is also considered to contain the point:

![image](https://github.com/copilot-emacs/copilot.el/assets/183/fd057e27-83b1-4919-ac11-91f45eb1b7da)

Notice:

* current binding for TAB is copilot-accept-completion, showing it recognizes the copilot-completion-map.
* highlighted eglot overlay, indicating that Emacs considers it to contain the current point.
* highlighted copilot overlay, indicating that Emacs also considers it it to contain the current point.

---

Screenshot showing that copilot-completion-map IS NOT active if another overlay contains the point, but copilot's overlay IS NOT considered to contain the point.

![image](https://github.com/copilot-emacs/copilot.el/assets/183/d6fffcee-a68e-4f0c-9350-fc54e89bd4ea)

Notice:

* current binding for TAB is indent-for-tab-command, ignoring the copilot completion map.
* highlighted eglot overlay, indicating that Emacs considers it to contain the current point.
* dimmed copilot overlay, indicating that Emacs does not consider it to contain the current point.

---

Screenshot showing that copilot-completion-map IS ACTIVE if the copilot overlay is active, but NO OVERLAY is considered to contain the point (notice key binding and dimmed-out color for the copilot overlay description).

[...]

PR #255, opened 2024-02-08
Use 1-character-long invisible overlay to activate copilot-completion-map

timcharper:
0-length overlays cannot contain the point, and if another overlay does contain the point, the keymap from the 0-length overlay will not be active.

As a workaround, we create two overlays for copilot completion: one to display the completion candidate, and another to activate the copilot-completion-map.


timcharper:
Closes #251

emil-vdw (approved):

Issue #258, opened 2024-02-12
jsonrpc process connection error


ezaquarii:
Trying to call any copilot function results in error:

```
Invalid slot name: "#<jsonrpc-process-connection jsonrpc-process-connection-1aa5a44>", :events-buffer-config
```

I'm not sure how to debug the problem. Installed using manual setup procedure.

jasalt:
There's multiple duplicate issues for "Invalid slot name" error resulting from jsonrpc package version mismatch as the built-in version is too old.

- https://github.com/copilot-emacs/copilot.el/issues/257
- https://github.com/copilot-emacs/copilot.el/issues/244
- https://github.com/copilot-emacs/copilot.el/issues/232

Also the readme was updated (https://github.com/copilot-emacs/copilot.el/commit/a06d27a24dd6a90ece41b29be743dd9710df19e8) to include mention about it in "Ensure your Emacs version is at least 27, the dependency package editorconfig (melpa) and jsonrpc (elpa, >= 1.0.24) are both installed.".

However having to manually "pin" packages over built-in package versions this way is not a very usual process and I think it should be better documented and included in a copy-paste starter config so that it would (most probably) work.

jasalt:
Had same problem and got it fixed by searching and selecting `jsonrpc` from `package-list-packages` and clicking install button there.

Couldn't figure out yet how to install it with `use-package` in config, being new to pinning packages like this. The up-to-date jsonrpc 1.0.24 package is in [GNU ELPA](https://elpa.gnu.org/packages/jsonrpc.html). Tried following that did not have effect:

```
(use-package jsonrpc :ensure t :pin gnu)  ; GNU ELPA install
```
There was also some weird network issue with that for a moment might have affected..

Package it's also mirrored in https://github.com/emacs-straight/jsonrpc/, so I tried install with straight:

```
(use-package jsonrpc
:straight (:type git :host github :repo "emacs-straight/jsonrpc" :files ("*" (:exclude ".git")))
:ensure t)
```

It cloned repo in `~/.emacs.d/straight/repos` and built it to `~/.emacs.d/straight/build`, but emacs doesn't seem to load those over built-in version and same error messages remained.

[...]

PR #259, opened 2024-02-12
fix: Better compatible with jsonrpc

jcs090218:
For #258.

### DIFF
--- a/copilot.el
+++ b/copilot.el
@@ -126,6 +126,17 @@ indentation offset."
   "List of buffers that have been opened in Copilot.")
 
 ;;
+;; jsonrpc
+;;
+
+(defun copilot--jsonrpc-version ()
+  "Return the jsonrpc version."
+  (let* ((desc (package-get-descriptor 'jsonrpc))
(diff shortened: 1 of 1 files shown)


emil-vdw:
Have you tested on old and new version of `jsonrpc`?

jcs090218:
Yes, it both works for me. :)

emil-vdw:
Don't you think that it would be better for this "old" `jsonrpc` version to be hardcoded? It's not really like it can change and we should only use the old parameter when it is older than that specific version not when it is older than the version we currently rely upon. I.e. at some point we may update the version of `jsonrpc` we depend on but the logic would break.

emil-vdw:
Nevermind, I was confused there for a second. LGTM. :+1:

jkl1337:
This is completely broken. Compilation breaks due to Package-Requires not at sufficient version (28.1). Furthermore, package-get-descriptor is only going to work if the descriptor is loaded first, and just requiring 'jsonrpc doesn't do that. As it is this breaks emacs startup.

PR #263, opened 2024-02-15
Find jsonrpc version by loading file directly

### LINKED ISSUE #262: PR #259 breaks startup
The commit in PR #259 has a number of problems and breaks emacs startup in most cases especially if using something like straight.el.

- `package-get-descriptor` is emacs 28.1 and above, so the Package-Requires should be updated. (Suggest using package-lint to avoid this).
- `(require 'package)` is missing so compilation breaks.

More troublesome is that `package-get-descriptor` only works if the descriptor is fetched and loaded, they do not need to be loaded normally to use/require a package. 
Is there a way to test for the needed features without pulling package metadata?

jkl1337:
Fix #262. Amendment to PR #260 with added require.

### DIFF
--- a/copilot.el
+++ b/copilot.el
@@ -12,6 +12,7 @@
 (require 'cl-lib)
 (require 'json)
 (require 'jsonrpc)
+(require 'package)
 (require 's)
 (require 'dash)
 (require 'editorconfig)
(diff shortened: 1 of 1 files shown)


jcs090218:
There is three issues with this PR.

1. `find-file-noselect` will open the file; we need to close it afterward.
2. From Pt. 1, if `jsonrpc.el` is opened by the user then don't close it.
3. It works on the installed version (install from ELPA), but It doesn't work on the built-in version (it returns `nil`).

PR #272, opened 2024-02-22
update version check to support lib64 directory

LemonBreezes:
Hi. On my Linux distribution, we use `lib64` instead of `lib` so I updated the version check.


jcs090218 (approved):
LGTM.

jcs090218:
Thank you for your contribution!

PR #274, opened 2024-02-24
feat: add: send minor mode language id to copilot

ncaq:
What
===

If there is a match for minor-mode before major-mode, it is sent to Copilot as a language id.

Why
===

Because git-commit-mode in git-commit.el is currently minor-mode and major-mode specifies text-mode, markdown-mode, etc. I felt that the accuracy was improved when this was correctly recognized as writing git commit messages.


jcs090218 (approved):
LGTM. :)

PR #281, opened 2024-03-13
Map caml-mode to OCaml

kevinji:
caml-mode is [an older major mode for OCaml](https://github.com/ocaml/caml-mode) that I forgot about in my previous PR.


jcs090218 (approved):
Thank you!

PR #298, opened 2024-04-28
docs: Provide example setup for use-package on Emacs 30+

tedroden:
Newer versions of `use-package` provide a `:vc` keyword. Users can now install copilot.el on standard emacs without `quelpa` or `straight`.

That version of `use-package` is already in Emacs 30.

Updated `README.md` to show an example.

(Thank you so much for this package!)

### DIFF
--- a/README.md
+++ b/README.md
@@ -126,18 +126,19 @@ dotspacemacs-additional-packages
 
 #### 1. Load `copilot.el`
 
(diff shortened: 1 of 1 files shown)


zerolfx:
image
GitHub doesn't support `#######`. Could you use bold or plain text instead?

jcs090218 (approved):

jcs090218:
Thanks for the contribution! :D

PR #299, opened 2024-04-29
feat: allow to customise arguments to copilot server

alexmurray:
When copilot is installed as a third-party package (say from https://snapcraft.io/copilot-client) it needs to be invoked with the "--stdio" argument - so provide a customisable variable which allows a user to specify such an argument - but set this to nil so it has no effect otherwise.


jcs090218 (approved):
LGTM.

PR #305, opened 2024-05-07
feat: Use the latest server

### LINKED ISSUE #304: update copilot-version
btw, new versions don't seem to use `bin/copilot-node-server` anymore.

jcs090218:
Fix #304.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -31,6 +31,9 @@ jobs:
         - os: macos-latest
           emacs-version: snapshot
           experimental: true
+        exclude:
+          - os: macos-latest
+            emacs-version: 27.2
 
     steps:
     - uses: actions/checkout@v4
--- a/copilot.el
+++ b/copilot.el
@@ -102,7 +102,6 @@ performance."
   :group 'copilot
   :type 'integer)
 
-
 (defcustom copilot-clear-overlay-ignore-commands nil
   "List of commands that should not clear the overlay when called."
   :group 'copilot
@@ -140,7 +139,7 @@ find indentation offset."
 (defconst copilot--server-executable
   (if (eq system-type 'windows-nt)
       (f-join copilot-install-dir "node_modules" "copilot-node-server"
-              "bin" "copilot-node-server")
+              "copilot" "dist" "agent.js")
     (f-join copilot-install-dir "lib" "node_modules" "copilot-node-server"
             "copilot" "dist" "agent.js"))
   "The dist directory containing agent.js file.")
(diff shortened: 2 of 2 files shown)


zerolfx (approved):

jidicula:
> it looks like a significant reworking of the protocol between the local nodejs server and the client (either in vim or emacs) happened in the 1.28.0 release - see [github/copilot.vim@f45ecd9](https://github.com/github/copilot.vim/commit/f45ecd95c1474885e278f20d6ae86a6e30de2151) for details - so I think copilot.el would need similar updates to work with this newer version

could the default be brought forward to 1.27.0 perhaps?

kassick:
Looking at the code of the vim plugin, it looks like the node server is a LSP server with support for `textDocument/inlineCompletion`. One option would be to let `lsp-mode` handle all the I/O (such as [here](https://github.com/kassick/gitlab-lsp.el/blob/main/gitlab-lsp.el#L380) and make the UI [simply interact with it](https://github.com/kassick/lsp-inline-completions)

Wrapping the `lsp-inline-completion` in a `with-lsp-workspace workspace` would ensure that only result from the target server are called:

```lisp
(defun copilot-inline-completions ()
  (interactive)

  (-when-let (workspace (--some (lsp-find-workspace it (buffer-file-name))
                               '(copilot-lsp copilot-lsp-remote)))

        (with-lsp-workspace workspace (lsp-inline-completion))))
```

I'll try to run the node server with lsp-mode and I'll let you know if it worked as expected

kassick:
Yup, it kinda works -- https://github.com/kassick/copilot-lsp

copilot-lsp + gitlab-lsp + lsp-inline-completions (plus an upstream fix to lsp-mode) displays the inline completions

jcs090218:
> could the default be brought forward to 1.27.0 perhaps?

Does it solve the issue if we used this version? 🤔

kassick:
> @jidicula Does it solve the issue if we used this version? 🤔

_the issue_ as in "it starts when using the latest version of the server"? I'm using version 1.37.0 -- it starts, it completes when asking for `inlineCompletions`, so, for this, it does.

But the solution in my repo goes in a rather different direction -- use lsp-mode to do the I/O, instead of using json-rpc. It lets all the typical issues (document sync, start/restart, callbacks after completion, etc.) to lsp-mode to handle.

One may try to backport the relevant portions of lsp-mode to copilot.el or go in the direction I went in the POC  -- I'm not certain which is best, TBH.

(it the issue is the reason why this MR was trying to upgrade to `>=1.30.0`, then I have no idea, sorry ;) )

Issue #300, opened 2024-04-29
copilot on Windows and replace-match make everything upper-case


ultronozm:
I was testing my config out on Windows with Emacs 29.3 and a fresh install of copilot.el, and encountered a strange bug.

Steps to reproduce: in any buffer where `copilot-mode` is enabled, use `search-forward` and `replace-match` to modify any string that begins with begins with an upper-case letter.  Then the following crazy thing happens: all text coming earlier in the same buffer has its capitalization modified to match that of the replaced string.

For example, if the buffer consists of the lines
```
foo
Bar
```
and we position point at the beginning and do `M-: (progn (search-forward "Bar") (replace-match "sup")), then the buffer looks like
```
Foo
Sup
```

This of course makes copilot unusable in conjunction with anything that uses `replace-match` nontrivially; the buffer quickly becomes fully capitalized.

I wasn't able to diagnose the issue via edebug; it doesn't help that `replace-match` is in the C source code.  The error doesn't seem to be with `copilot--on-doc-change` and `copilot--post-command-hook`.  The offending substitution happens before the call to `copilot--post-command-hook` and between the second and third calls to `copilot--on-doc-change` during the edit.

Any advice for debugging this would be most appreciated.

PR #309, opened 2024-05-20
fix: weird capitalization bug (#300)

ultronozm:
This patch saves the match data inside `copilot--on-doc-change`, fixing a bizarre bug that I encountered only on Windows (#300).  Compare with the second paragraph of https://www.gnu.org/software/emacs/manual/html_node/elisp/Change-Hooks.html.


ultronozm:
Please let me know if I've misunderstood, but the test failure doesn't seem like it's my fault.

zerolfx (approved):

zerolfx:
This closes #300 and #323

PR #311, opened 2024-06-01
Add customization to disable max char warning

Gleek:
Completions work well enough for me even when the file is truncated. Getting a warning every time is therefore unecessary.

### DIFF
--- a/copilot.el
+++ b/copilot.el
@@ -116,6 +116,11 @@ find indentation offset."
   :group 'copilot
   :type 'boolean)
 
+(defcustom copilot-max-char-warning-disable nil
+  "Disable maximum characters warning."
+  :group 'copilot
+  :type 'boolean)
+
 (defcustom copilot-indentation-alist
   (append '((latex-mode tex-indent-basic)
             (nxml-mode nxml-child-indent)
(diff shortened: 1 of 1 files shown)


amake:
This has been suggested and rejected in #273.

However I also came here with the intent to ask for this sort of feature. My problem is that I get warnings like the below when I update packages via e.g. `package-update-all`. 

```
⛔ Warning (copilot):  *temp*-529187 size exceeds 'copilot-max-char' (100000), copilot completions may not work
```

I am activating `copilot-mode` in `prog-mode-hook`, so I assume that it's catching large Elisp files that are opened at package install time. I *do* want `copilot-mode` in regular Elisp files, so I don't think it's feasible to be more selective about what hook I put it on.

Would it make sense to have copilot ignore buffers with the `*temp*-` prefix?

Gleek:
Got it. I missed that this has already been addressed. I'll close this here. Thanks for pointing out.

I think you could disable copilot for such buffers, as @emil-vdw suggested in the original issue thread. Maybe you could modify it slightly to handle your case specifically. Something like:

```elisp
(defun my/activate-copilot ()
  (if (or (> (buffer-size) copilot-max-char)
          (string-prefix-p "*temp*-" (buffer-name)))
      ;; Or don't even warn to get rid of it.
      (warn "Buffer size exceeds copilot max char limit or buffer is temporary. Copilot will not be activated.")
    (copilot-mode)))
```

amake:
My position is that I shouldn't have to concoct my own custom activation mechanism to avoid warnings during normal Emacs operations like updating a package. I think that's something that copilot should handle out of the box.

Gleek:
Understand what you're saying now. I suppose this could be discussed in a dedicated issue of it's own.

PR #319, opened 2024-06-22
Verify buffer is present when running the success func

Gleek:
This is a fix for the issue where copilot is triggered in temporary buffer that could get killed before the success function is called. Eg: Capture, commit-editmsg buffers etc.


LemonBreezes:
I am having this issue as well so it would be great if we could have this merged.

LemonBreezes:
@jcs090218

zerolfx (approved):

PR #325, opened 2024-07-16
Indentation docs

iamkarlson:
It seems like this is not clear on how to configure `copilot-indentation-alist`, and for the new doom emacs users it can be twice as confusing


jcs090218 (approved):

jcs090218 (approved):
Thanks!

PR #334, opened 2024-08-15
refactor: Consolidate copilot server executable path logic

LemonBreezes:
Hello. I am using Gentoo Linux and on my machine the path to the `agent.js` file is a little bit different than what is currently coded so I have made this PR to fix the detection of `agent.js`.


jcs090218 (approved):
LGTM.

zerolfx (approved):

PR #335, opened 2024-08-30
fix: use `standard-indent` instead of `tab-width`

ncaq:
I see the following warnings.

```
Warning: copilot--infer-indentation-offset found no mode-specific indentation offset, using ’tab-width’ instead.  You can suppress this error message by customizing ’copilot-indent-warning-suppress’.
```

Reading the documentation comments for `tab-width` shows that it has no effect if spaces are used for indentation.

```
Documentation
Distance between tab stops (for display of tab characters), in columns.

This controls the width of a TAB character on display.
The value should be a positive integer.
Note that this variable doesn't necessarily affect the size of the
indentation step.  However, if the major mode's indentation facility
inserts one or more TAB characters, this variable will affect the
indentation step as well, even if indent-tabs-mode is non-nil.
```

If tabs are used, the `tab-width` is automatically adjusted, so there seems to be no need to worry about it.

Conversely, `tabSize` is supposed to be the real tab size, not the indentation size, so put the value as it is.

### DIFF
--- a/copilot.el
+++ b/copilot.el
@@ -484,7 +484,7 @@ automatically, browse to %s." user-code verification-uri))
           (display-warning '(copilot copilot-no-mode-indent)
                            "copilot--infer-indentation-offset found no mode-specific indentation offset.")
           (setq-local copilot--indent-warning-printed-p t))
-        tab-width)))
+        standard-indent)))
 
 (defun copilot--get-relative-path ()
   "Get relative path to current buffer."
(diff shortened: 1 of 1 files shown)


jcs090218 (approved):

jcs090218:
Make sense! Thanks!

ncaq:
I looked back and realized that I should change the `:tabSize` in `copilot--generate-doc` to `tab-width`.

ncaq:
I rewrited.

jcs090218 (approved):

jcs090218:
Thank you!

PR #338, opened 2024-09-07
feat(version): Upgrade to Copilot 1.40.0

knilink:
update lsp to 1.40
- added `initialized` notification which required by later lsp server
- updated script entry from `agent.js` to `language-server.js`, assuming not supporting backward compatible since copilot.el enforcing `copilot-node-server`'s version

i am not subscribing copilot and only tested with modded lsp server based on [copilot.vim](https://github.com/github/copilot.vim) so may need someone else help testing it

P.S. 
proposed `copilot-node-server` to add executable entry point https://github.com/jfcherng/copilot-node-server/issues/17
so start from next release the server should be easily started by `npx -y  copilot-node-server@1.40.0 --stdio` therefor can consider simplify the installation and version management later

### DIFF
--- a/copilot.el
+++ b/copilot.el
@@ -140,9 +140,9 @@ find indentation offset."
   :group 'copilot)
 
 (defvar copilot--server-executable nil
-  "The dist directory containing agent.js file.")
+  "The dist directory containing language-server.js file.")
 
-(defcustom copilot-version "1.27.0"
+(defcustom copilot-version "1.40.0"
   "Copilot version.
 
 The default value is the preferred version and ensures functionality.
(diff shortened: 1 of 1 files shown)


alexmurray:
FWIW (I am not part of the copilot-emacs project but just an avid user), the upstream copilot.vim project renamed the agent.js file to language-server.js in the 1.34.0 release, so you will also need to update the references to this as well in `copilot--server-executable` - but otherwise I tested this briefly and it works for me! Thanks.

jfcherng:
> the upstream copilot.vim project renamed the agent.js file to language-server.js in the 1.34.0 release

Yes, this kind of issue can be avoid by @knilink's proposal. It can be directly dealt in https://github.com/jfcherng/copilot-node-server if the user uses `npx` to start the server.

patzy:
I had a problem with copilot.el not working  on some file.
Enabling logs I had the following message: `Copilot is disabled because we could not fetch the repository policy`
After tinkering it appears that being in a git tracked directory triggered the problem.

Not sure what the issue is but upgrading with this PR seems to fix the issue.
Thanks

rakotomandimby:
How does it deal with #346 ?

kiennq:
> How does it deal with #346 ?

I should resolve that issue.

bbatsov:
Seems https://github.com/copilot-emacs/copilot.el/pull/363 rendered this obsolete.

PR #339, opened 2024-09-07
feat: customizable lsp event handler and lsp server log

knilink:
- adding custom even handler registration function `copilot-on-notification`, `copilot-on-request`
- adding `window/logMessage` handler and buffer `*copilot-language-server-log*`  for lsp server logging

image

PR #343, opened 2024-09-25
copilot chat poc

knilink:
[chrome-extension-nlipoenfbbikpbjkfpfillcgkoblgpmj-setup-react-html-from-install.webm](https://github.com/user-attachments/assets/67ee7bde-ecd6-48f5-97f1-2a68b5087d79)

just to share some findings about how conversation feature can possibly be implemented for anyone who is interested

language server will behave in 2 different way depends on "source" parameter provided in the `conversation/create` request:
- `panel`: server will analysis the conversation at the end of the response and will suggest title and follow up question
- `inline`: server will try to update the code and notify editor about the new file content, may need to select a region otherwise it's assumed the entire filed is selected and the llm may be confused about what it needs to do

the server use `$/progress` notification to send all kinds of responses
the server send `conversation/context` request to collected chat context, if not handled, server will assume no filed is opened

### DIFF
--- a/README.md
+++ b/README.md
@@ -296,6 +296,27 @@ For example:
 (setq copilot-network-proxy '(:host "127.0.0.1" :port 7890))
 ```
 
+### copilot-on-request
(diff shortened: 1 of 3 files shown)


bbatsov:
Let's revisit this PR after #339 gets merged.

PR #346, opened 2024-10-30
Support find language-server.js file since copilot-node-server rename agent.js to language-server.js

eval-exec:
copilot-node-server renamed agent.js to language-server.js:
https://github.com/jfcherng/copilot-node-server/pull/10/files


bbatsov:
I think this can be closed given https://github.com/copilot-emacs/copilot.el/pull/363 got merged.

Issue #313, opened 2024-06-04
Don't warn about copilot-max-char during normal Emacs functions


amake:
I get warnings like the below when I update packages via e.g. `package-update-all`. 

```
⛔ Warning (copilot):  *temp*-529187 size exceeds 'copilot-max-char' (100000), copilot completions may not work
```

I am activating `copilot-mode` in `prog-mode-hook`, so I assume that it's catching large Elisp files that are opened at package install time. I *do* want `copilot-mode` in regular Elisp files, so I don't think it's feasible to be more selective about what hook I put it on.

Would it make sense to have copilot ignore buffers with the `*temp*-` prefix?

_Originally posted by @amake in https://github.com/copilot-emacs/copilot.el/issues/311#issuecomment-2146427990_

ryankhart:
I had the same issue. And thanks to that pull request that was made to, at least partially, resolve this, managed to suppress the warning by adding this line to my init.el (if you're using Doom Emacs, thought, it would be in your config.el)
```Elisp
(setq copilot-max-char-warning-disabled t)
```

You might have already known and done this, but for anyone else searching for solutions to this issue and is even newer to Emacs than I am, hopefully, this helps.

jackgris:
Hi! I still see this warning even after following the advice:

> I had the same issue. And thanks to that pull request that was made to, at least partially, resolve this, managed to suppress the warning by adding this line to my init.el (if you're using Doom Emacs, though, it would be in your config.el)
> 
> (setq copilot-max-char-warning-disabled t)
> 
> You might have already known and done this, but for anyone else searching for solutions to this issue and is even newer to Emacs than I am, hopefully, this helps.

Is anyone else still experiencing the same issue?

Also, just to clarify, could the variable name be without the final d, like this?
```lisp
(setq copilot-max-char-warning-disable t)
```
Lastly, even when I set:
```lisp
(setq copilot-max-char -1) ;; -1 means no limit
```
I'm still getting the same warning in Spacemacs. I confirmed that both variables are set to the correct values.

jackgris:
At least for me, the only solution I’ve found is to modify the value directly in the function in `copilot.el` to a higher number or to `-1`.
```lisp
(defcustom copilot-max-char 1000000
  "Maximum number of characters to send to Copilot, -1 means no limit."
  :group 'copilot
  :type 'integer
  :package-version '(copilot . "0.1"))
```

Or `copilot-max-char-warning-disable` to `true`
```lisp
[...]

PR #349, opened 2024-11-19
feat: adding copilot-max-char-warning-disable variable

tninja:
During editing large org file or source code file, the warning could be annoying.

Added a variable to control the msg go to warning, or mini buffer (message)

### DIFF
--- a/copilot.el
+++ b/copilot.el
@@ -116,6 +116,11 @@ find indentation offset."
   :group 'copilot
   :type 'boolean)
 
+(defcustom copilot-max-char-warning-disable nil
+  "When non-nil, disable warning about buffer size exceeding `copilot-max-char'."
+  :group 'copilot 
+  :type 'boolean)
+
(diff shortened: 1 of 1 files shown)


jcs090218:
I have no objection to this, but I might let other maintainers decide whether this should be merged. 🤔

AlexChesters:
just chiming in to say that this feature could be a useful solution for the problem described in #313

mrcnski:
Cool! Is anything blocking this PR from being merged? I independently requested this feature, before finding this PR:

https://github.com/copilot-emacs/copilot.el/issues/175#issuecomment-2571724404

> Is there a way to suppress the warning, or make it a message instead? For me, the issue's not critical enough to appear in the `*Warnings*` buffer. For now I'll just bump the limit. :)

jcs090218 (approved):

jcs090218:
Merged! Thanks to everyone who participated.

PR #350, opened 2024-12-12
Address MELPA comments (update summary and use derived-mode-p)

LaurenceWarne:
Hi, I saw there were a few outstanding issues on https://github.com/melpa/melpa/pull/9217 / https://github.com/copilot-emacs/copilot.el/issues/120 and they seemed easy enough to address except for: ```You should depend on (emacs "29.1") if you need `python-ts-mode'.``` which as suggested by the melpa maintainer, seems like a false positive to me.

Hope this is helpful, thanks!


jcs090218 (approved):

DamienCassou (approved):
This looks good to me

jcs090218:
Thank you!

PR #351, opened 2024-12-21
Changing and fixing indentation handling

### LINKED ISSUE #28: Visual cursor jumps to end of completion while real position stays put
Ever since commit dfa803e757d357b08d1702093abd1362287c1e73 I'm afraid I've had issues with the cursor position.

Physically, the position does not change, and any cursor movement/editing actions briefly show the cursor in the correct position again, until a copilot suggestion appears. However it only happens for single-line suggestions from copilot. If it's suggesting multiple lines, the visual cursor position stays in the right place.

Here's a couple of gifs:

On commit 56b7d1b4cff6b966ca7125b1678438097ea63050 (latest):

copilot-current

On commit 3dcf2a74eaa3d08c3cf30b3b24ac93ae4b40be90 with `copilot-overlay-safe` set to `nil`:

copilot-old

aasseman:
Correct for indentation after accepting a completion rather than during typing. This is less confusing than being impeded while typing spaces or tabs on a new line while there is completion preview, and behaves more like VSCode.

Fixes #28 and #200.

### DIFF
--- a/copilot.el
+++ b/copilot.el
@@ -759,14 +759,12 @@ To work around posn problems with after-string property.")
 already saving an excursion.  This is also a private function."
   (copilot-clear-overlay)
   (when (and (s-present-p completion)
-             (or (= start (point))      ; up-to-date completion
-                 (and (< start (point)) ; special case for removing indentation
-                      (s-blank-p (s-trim (buffer-substring-no-properties start (point)))))))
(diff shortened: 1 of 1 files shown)


jcs090218 (approved):

jcs090218:
Thank you!

PR #354, opened 2024-12-24
Fix #352

### LINKED ISSUE #352: Copilot completions are inserted before the point where suggestion begin.
Using Emacs 29.4 on Void Linux with Evil, and installed `copilot.el` with [Elpaca](https://github.com/progfolio/elpaca) couple hours ago.
Not sure how to describe this behavior in words, so here is the video of what is going on.
Please do change the title if there are better ways to express this.

https://github.com/user-attachments/assets/3c5ee180-26b2-41ae-8c89-8d86d97c93c8

As you can see, when I type the whole "completion candidate" myself, the last character is inserted not at the end of the completion (where it should be), but where the completion candidate was presented originally.

In the first example, I invoked `copilot-accept-completion` by pressing `<tab>`. It seems to work fine on simple cases, but I've seen the previous character getting "eaten" and spat out at the end of the completion from time to time. I cannot reproduce this reliably, though.

This package was working fine before, and it seems today's reinstallation has introduced this behavior.

Here is the relevant section in my configuration file: https://github.com/RangHo/dotfiles/blob/main/emacs/.config/emacs/config.org#github-copilot

If you need any more info, please do let me know!

aasseman:
I hope this will fix the issue I introduced. From my local tests it does fix the issue.

Fixes #352

### DIFF
--- a/copilot.el
+++ b/copilot.el
@@ -760,11 +760,11 @@ already saving an excursion.  This is also a private function."
   (copilot-clear-overlay)
   (when (and (s-present-p completion)
              (or (<= start (point))))
-  (let* ((ov (copilot--get-overlay)))
+    (let* ((ov (copilot--get-overlay)))
       (overlay-put ov 'tail-length (- (line-end-position) end))
       (copilot--set-overlay-text ov completion)
       (overlay-put ov 'uuid uuid)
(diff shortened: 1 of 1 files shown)


jcs090218 (approved):

jcs090218:
Thank you!

PR #357, opened 2024-12-24
Fix #352 again

### LINKED ISSUE #352: Copilot completions are inserted before the point where suggestion begin.

aasseman:
I'll need some help with this. This fixes the issue observed in https://github.com/copilot-emacs/copilot.el/issues/352#issuecomment-2560585257, but I'm not sure this is the correct way and if it will create other problems.

The first observation is that the code deletion problem is not coming from the the code I added that removes extra indentation. So I don't fully understand why my changes created the issue.
I also observed that `start` and `end` would always be the same value, unless some text gets inserted before the completion (repro of the issue). So setting `start` to `end` "fixes" it.

The problem is that I don't fully understand what the purpose of those 2 values is. Is it for a refactor/rewrite feature?


rakotomandimby (approved):

guillaumebrunerie:
I think this should be reverted for now, it results in other issues (#405, #409) that I think are a lot more annoying that the original issue it is trying to fix. In particular #409 makes it impossible to just write code normally with `copilot-mode` enabled: as soon as it correctly guesses the last character of a line, it gets mangled. It happens all the time to me with semicolons.

bbatsov:
Yeah, I'll revert it. Would have done it earlier, if I could just commit this directly, but now it's a bit of a hassle.

PR #358, opened 2024-12-27
Improve error message

DamienCassou:
Instead of telling the user that a newer version exists even when the
user is using the latest, the new error message explains what is
expected and why.


jcs090218 (approved):

jcs090218:
LGTM.

jcs090218:
Thank you!

PR #363, opened 2025-02-16
Support official Github copilot LSP and allow to install the latest LSP version

kiennq:
Support official Github copilot LSP and allow to install the latest LSP version via `copilot-version = nil`.

### DIFF
--- a/copilot.el
+++ b/copilot.el
@@ -97,11 +97,6 @@ performance."
   :group 'copilot
   :type 'integer)
 
-(defcustom copilot-node-executable (executable-find "node")
-  "Node executable path."
-  :group 'copilot
-  :type 'string)
-
 (defcustom copilot-server-args '("--stdio")
   "Additional arguments to pass to the Copilot server."
   :group 'copilot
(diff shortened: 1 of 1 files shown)


eval-exec (approved):

jcs090218 (approved):

zerolfx (approved):

Kaylebor:
I've been trying to have copilot reference a local install of the LSP, like this on my config.el:

```lisp
(setq copilot-server-executable (expand-file-name "~/.bun/bin/copilot-language-server"))
```

The variable is correctly set when checking its value, but copilot.el complains that there's no server installed.
Is the only supported method to use the package-installed LSP, and never a globally-available one? With or without this variable, it fails to find it.

kiennq:
> I've been trying to have copilot reference a local install of the LSP, like this on my config.el (Doom Emacs):
> 
> ```commonlisp
> (after! copilot
>   (setq copilot-server-executable (expand-file-name "~/.bun/bin/copilot-language-server")))
> ```
> 
> I have also tried:
> 
> ```commonlisp
> (use-package! copilot
>   :after copilot
>   :config (setq copilot-server-executable (expand-file-name "~/.bun/bin/copilot-language-server")))
> ```
> 
> The variable is correctly set when checking its value, but copilot.el complains that there's no server installed. Is the only supported method to use the package-installed LSP, and never a globally-available one? With or without this variable, it fails to find it.

Should be fixed in #364

PR #369, opened 2025-02-21
copilot-server-executable: support executable from path

terlar:
Since inside `copilot--start-agent` it does a `file-exists-p`-check we need to return the full path.


zerolfx:
fixed by #370

PR #380, opened 2025-03-04
Allow users to specify `settings` of Copilot LSP

jmpunkt:
## Feature
This allows users to set the `settings` (https://github.com/github/copilot-language-server-release?tab=readme-ov-file#configuration-management) for the Copilot LSP. VSCode stores this in `settings.json`? Having this feature allows users to login into GitHub Enterprise instances. Without this it is not possible. You can also disable telemetry with this.

## How To Test
Testing this without a GHE instance is tricky but doable.

1. Start Emacs and evaluate this PR.
2. Log out of your current session (`M-x copilot-logout`).
3.  Use `eval-last-sexp (C-x C-e)` to step through the following snippets

```elisp
(copilot--start-agent)
(copilot-login)
;; you should be able to login, everything works fine (abort login C-g)
(setopt
 copilot-lsp-settings
 '(:github-enterprise (:uri "https://example2.ghe.com")))
(copilot-login)
;; there should be an error, the instance does not exist, the error code shows the url
```
Another test can be:
```elisp
(setq copilot-log-max 2000)
(copilot--start-agent)
(switch-to-buffer-other-window "*copilot events*")
(setopt
 copilot-lsp-settings
 '(:github-enterprise (:uri "https://example2.ghe.com")))
;; you will see in the log that the telemetry tries to fetch from the instance (with an error of course)
(setopt
 copilot-lsp-settings
 '(:github-enterprise (:uri "https://example2.ghe.com")
   :telemetry (:telemetryLevel "all")))
;; now the previous telemetry error is gone
```

### DIFF
--- a/copilot.el
+++ b/copilot.el
@@ -178,11 +178,21 @@ You may adjust this variable at your own risk."
   :package-version '(copilot . "0.1"))
 
 (defun copilot--lsp-configuration-changed (symbol value)
-  "Notify the Copilot LSP configuration has changed and set SYMBOL to VALUE."
+  "Restart the Copilot LSP due to SYMBOL changed to VALUE.
+
+This function will be called by the customization framework when the
+`copilot-lsp-settings' is changed.  When changed with `setq', then this function
+will not be called."
   (let ((was-bound (boundp symbol)))
     (set-default symbol value)
     (when was-bound
-      (copilot--notify-configuration))))
+      ;; Notifying the agent with the new value does only work if we include the
+      ;; last value (as nil) as well. For example, having the value
+      ;; '(:github-enterprise (:uri "https://example2.ghe.com")) and setting it
+      ;; to nil would require to send the value '(:github-enterprise (:uri nil))
+      ;; to the server. Otherwise, the value is ignored, since sending nil is
+      ;; not enough.
+      (copilot--start-agent))))
 
 (defcustom copilot-lsp-settings nil
   "Settings for the Copilot LSP server.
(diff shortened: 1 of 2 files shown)


bbatsov:
I think you should also mention this in the README.

jmpunkt:
I added the README entry. If there is something else missing, then let me know.

copilot.el
```diff
@@ -420,6 +443,13 @@ SUCCESS-FN is the CALLBACK."
        ;; handle older jsonrpc versions
        (funcall make-fn :events-buffer-scrollback-size copilot-log-max)))))
 
+(defun copilot--notify-configuration ()
```
bbatsov, line 446: I think "copilot--notify-configuration-changed" is probably a better name for this function.

copilot.el
```diff
@@ -177,6 +177,29 @@ You may adjust this variable at your own risk."
...
+https://github.com/github/copilot-language-server-release?tab=readme-ov-file#configuration-management
+for complete documentation.
+
+For example to use GitHub Enterprise use the following configuration:
+ '(:github-enterprise (:uri \"https://example.ghe.com\"))
+
+Exchange the URI with the correct URI of your organization."
+  :set #'copilot--lsp-configuration-changed
+  :type 'sexp
+  :group 'copilot)
```
bbatsov, line 201: Add a `:package-version` property mentioning this got added in version 0.2.

copilot.el
```diff
@@ -177,6 +177,29 @@ You may adjust this variable at your own risk."
...
+This value will always be sent to the server when the server starts or the value
+changes.  See
+https://github.com/github/copilot-language-server-release?tab=readme-ov-file#configuration-management
+for complete documentation.
+
+For example to use GitHub Enterprise use the following configuration:
+ '(:github-enterprise (:uri \"https://example.ghe.com\"))
+
+Exchange the URI with the correct URI of your organization."
+  :set #'copilot--lsp-configuration-changed
```
bbatsov, line 199: I'm reasonably certain this will be triggered only when using `M-x customize`, so you'll probably have to mention it somewhere, so people are not surprised that manual config changes don't trigger it.

copilot.el
```diff
@@ -177,6 +177,29 @@ You may adjust this variable at your own risk."
   :group 'copilot
   :package-version '(copilot . "0.1"))
 
+(defun copilot--lsp-configuration-changed (symbol value)
+  "Notify the Copilot LSP configuration has changed and set SYMBOL to VALUE."
```
bbatsov, line 181: You should probably add a note that's supposed to get triggered when `copilot-lsp-settings` gets updated via the customize UI. Such callbacks are rarely used and might surprise many users.

bbatsov:
I've added a few small remarks here and there, but overall your changes are in a good shape.

copilot.el
jmpunkt: This function got removed. We only notify the server at the start now.

jmpunkt:
Thanks for the review. I had to change one thing. There is an issue with the notify when changing the settings from something (e.g. `'(:github-enterprise (:uri "test.com"))`) to `nil`. The change is not recognized by the LSP. We would have to send `'(:github-enterprise (:uri nil))` instead, in order to apply the change. So I restart the agent every time the settings change. Otherwise, we would have to calculate the difference between the old and new value.

bbatsov (approved):

bbatsov:
Great work! Thanks!

PR #392, opened 2025-03-30
Trim external deps

bbatsov:
The package uses a few external deps that are not really needed on modern Emacs:

- `s.el` is easily replaced with built-in functions
- `dash` is easily replaced with the built-in `seq`
- `f.el` can also be replaced easily with built-in functions

When it comes to external deps, less is always more. :-)

### DIFF
--- a/Eask
+++ b/Eask
@@ -13,8 +13,6 @@
 (source 'melpa)
 
 (depends-on "emacs" "27.2")
-(depends-on "s")
-(depends-on "dash")
 (depends-on "editorconfig")
 (depends-on "jsonrpc")
 (depends-on "f")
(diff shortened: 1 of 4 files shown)


bbatsov:
I've dropped s.el and dash.el, and there's only `f-join` left from `f.el`. Probably it'd be best to just inline it, as it's a pretty simple function, but we can do this down the road.

bbatsov:
@zerolfx Seems I've made a silly typo. Just fixed it.

jcs090218 (approved):

viktorbezdek:
@bbatsov I guess the love for well written code never fades 🤘🏻

Hope you're doing great.

🫡

Issue #396, opened 2025-03-31
Error running timer: (setting-constant :text)


danp:
Hi there,

Just updated to latest (463ed62b7b00729118e0ab9f1588eea6f6b8e18d) from 7d105d708a23d16cdfd5240500be8bb02f95a46e and now getting this when I would expect to get completions:

```
Error running timer: (setting-constant :text)
```

If I revert back to 7d105d708a23d16cdfd5240500be8bb02f95a46e things resume working. Using Emacs 30.1 from https://github.com/d12frosted/homebrew-emacs-plus on macOS.

ckruse:
I get a variant of this error: `Error running timer: (wrong-type-argument number-or-marker-p nil)`

This is the backtrace:

```
Debugger entered--Lisp error: (wrong-type-argument number-or-marker-p nil)
  copilot--show-completion((:uuid "331118ae-013c-4dde-a6e4-d1a2eee8e243" :text "    <Route path=\"/confirm-event\" element={<ConfirmEvent />} />" :range (:start (:line 4 :character 0) :end (:line 4 :character 11)) :displayText "<Route path=\"/confirm-event\" element={<ConfirmEvent />} />" :position (:line 4 :character 11) :docVersion 112))
  #f(compiled-function (&rest rest) #<bytecode -0x1491f50896a20bab>)(:completions [(:uuid "331118ae-013c-4dde-a6e4-d1a2eee8e243" :text "    <Route path=\"/confirm-event\" element={<ConfirmEvent />} />" :range (:start (:line 4 :character 0) :end (:line 4 :character 11)) :displayText "<Route path=\"/confirm-event\" element={<ConfirmEvent />} />" :position (:line 4 :character 11) :docVersion 112)])
  apply(#f(compiled-function (&rest rest) #<bytecode -0x1491f50896a20bab>) (:completions [(:uuid "331118ae-013c-4dde-a6e4-d1a2eee8e243" :text "    <Route path=\"/confirm-event\" element={<ConfirmEvent />} />" :range (:start (:line 4 :character 0) :end (:line 4 :character 11)) :displayText "<Route path=\"/confirm-event\" element={<ConfirmEvent />} />" :position (:line 4 :character 11) :docVersion 112)]))
[...]

PR #397, opened 2025-04-01
Fix usages of `copilot--dbind`

DamienCassou:


bbatsov:
@zerolfx I'm reasonably sure this is also the fix for #396, so I think we should merge it ASAP.

zerolfx (approved):

zerolfx:
@DamienCassou Thanks a lot for this hot fix!

bbatsov:
Same! I had recently updated the `copilot--dbind` to remove the need for using `dash.el`, but I forgot to update its usages... 🤦‍♂️

PR #424, opened 2025-06-22
Fix buffer boundary handling in copilot--get-source

ultronozm:
This PR adds some boundary checks, fixing sporadic issues that I
encountered, and also adds a call to "widen" to fix the following
easily reproduced issue: copilot-mode does not work correctly when
activated in an already narrowed buffer.

* copilot.el (copilot--get-source): Widen at beginning.  Add buffer
boundary checks when truncating the buffer content.  Ensure that
substring operations never exceed buffer limits by clamping to valid
ranges.

### DIFF
--- a/copilot.el
+++ b/copilot.el
@@ -651,35 +651,39 @@ automatically, browse to %s." user-code verification-uri))
 
 (defun copilot--get-source ()
   "Get source code from current buffer."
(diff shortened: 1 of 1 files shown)


bbatsov (approved):
Looks good to me. Thanks!

Issue #422, opened 2025-06-17
Cannot get copilot to use 4 spaces to indent Python


kynnjo:
I have run all of the following:

```
    (setq python-indent-offset 4)
    (add-to-list 'copilot-indentation-alist '(prog-mode 4))
    (add-to-list 'copilot-indentation-alist '(python-mode 4))
    (setq-default indent-tabs-mode nil)
```

(My `tab-width` variable is set to 8, and that is where I want to keep it.  Since I don't want TABs in my Python code, at all, this setting should not matter.)

To no avail: all the code that Copilot proposes still uses an 8-space indentation.

What else must I do to get Copilot to use 4-spaces/level for indenting Python code?

kynnjo:
@bbatsov :  Yes, if I turn off `copilot-mode`, Python indentation works fine.  (I.e., I hit tab to increase/decrease indentation by 4 spaces at a time.)

Regarding your second question, I don't have editorconfig installed.  Is that what you where referring to?

kynnjo:
@masonkatz: It looks like, somehow, `copilot-mode` is using `tab-width` for indenting code.

I really don't like the idea of changing `tab-width`, though, since it affects more than code indentation.

The question is: why is `copilot-mode` ignoring its own `indentSize`?

kynnjo:
@masonkatz: Thank you!  It looks like, somehow, `copilot-mode` is using `tab-width` for indenting code.

I really don't like the idea of changing `tab-width`, though, since it affects more than code indentation.

The question is: why is `copilot-mode` ignoring its own `indentSize`?

bbatsov:
Yeah, it seems we'll need to fix this. PR welcome!

PR #425, opened 2025-06-27
fix: infer tabSize sent in the LSP formattingOptions (#422)

masonkatz:
Full discussion in issue #422 

- go back to inferring the `tabSize`
- keep `indentSize` unchanged

I've been running this for several days in Go and Python.


bbatsov:
Didn't you also say that indentSize is redundant?

masonkatz:
I believe that `indentSize` is not used, but may have been at one time.

- `tabSize` can be found as a variable in the language server .js code (also in the .map file).
- `indentSize` does not exist in the server .js code -- but it is in the .map file.

- While it is not in the copilot.vim code it does appear in the [xcode](https://github.com/github/CopilotForXcode) (also from github) version.

To be conservative I kept it in.

Which makes more sense?

1) take it out
2) add a comment saying it is suspected to be unused

bbatsov:
Add a comment so we won't forget to investigate this further down the road.

masonkatz:
PR updated

bbatsov (approved):

PR #429, opened 2025-09-15
Drop s library dependency

midsbie:
The function `s-equals-p` from the s library was referenced in the code but the dependency had not been declared.

Since the referenced function is nothing but a wrapper around `string-equal` (see below), this PR refactors the code to call `string-equal` directly and drop the dependency altogether.

```elisp
  (defun s-equals? (s1 s2)
    "Is S1 equal to S2?

  This is a simple wrapper around the built-in `string-equal'."
    (declare (pure t) (side-effect-free t))
    (string-equal s1 s2))
```


bbatsov:
>  Is there any specific reason why the files copilot-autoloads.el and copilot-pkg.el haven't been added as git ignore clauses? Can add, if no reason; just let me know.

Just an oversight.

PR #431, opened 2025-10-28
Replace manual change tracking with track-changes library

### LINKED ISSUE #430: Synchronization issue with complex edits
After using the commands `upcase-word` or `downcase-word` in a buffer with `copilot-mode` enabled, the completions are subsequently corrupted.  The issue is that copilot.el tacitly assumes that the range supplied to `before-change-functions` is precisely the text that will be replaced, but this assumption is false for some commands.

I've been testing on my local fork two different fixes for this issue, and would like feedback on which to upstream.  The first patches the issue directly by introducing and keeping track of some "pre-change" variables.  The second replaces the existing approach based on before/after change functions with use of the `track-changes` library (available on ELPA, and included in Emacs 30.1+).  I prefer the second because it seems more robust, but figured I'd ask for feedback before polishing it further.

ultronozm:
Fixes #430.

* copilot.el (track-changes): New requirement. (copilot--lsp-pos): New helper function.
(copilot--generate-doc): Use it.
(copilot--on-doc-change): Remove.
(copilot--track-changes-id): New buffer-local variable. (copilot--lsp-range-end-from-oldtext): New helper function. (copilot--track-changes-signal): New function to fetch and forward buffer changes.
(copilot--mode-setup): Remove old before-change-functions and after-change-functions hooks, register with track-changes instead. (copilot--mode-teardown): Remove old cleanup code, unregister from track-changes.

### DIFF
--- a/Eask
+++ b/Eask
@@ -18,5 +18,6 @@
 (depends-on "editorconfig")
 (depends-on "jsonrpc")
 (depends-on "f")
+(depends-on "track-changes" "1.4")
 
 (setq network-security-level 'low)  ; see https://github.com/jcs090218/setup-emacs-windows/issues/156#issuecomment-932956432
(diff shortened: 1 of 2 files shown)


bbatsov (approved):
Nice improvement! I hadn't noticed the `track-changes` library so far, but it looks quite handy.

PR #435, opened 2025-12-08
Honor ignored prefixes when clearing overlay (fixes #434)

### LINKED ISSUE #434: Some 'ignored' commands still clear the overlay
To reproduce, add any of the following commands to `copilot-clear-overlay-ignore-commands`:

- `universal-argument`, `universal-argument-more`, `digit-argument` `negative-argument`

Enable copilot, trigger a completion, and press, e.g., `C-u`.  Observe that the overlay disappears.

I'll send a PR that fixes this in a minute.  The practical gain is that we can now use prefix arguments like `C-3` with commands like `copilot-accept-completion-by-word`.

ultronozm:
Prefix commands like C-u were clearing the Copilot overlay even when
their command symbols (e.g., 'universal-argument) were listed in
'copilot-clear-overlay-ignore-commands'.

* copilot.el (copilot--post-command): Check both 'this-command and the
command bound to the current key sequence.

* test/copilot-overlay-tests.el
(copilot-post-command-honors-key-binding-ignore)
(copilot-post-command-clears-when-not-ignored): New tests.

* Eask: Add script entry to run the tests.

### DIFF
--- a/copilot.el
+++ b/copilot.el
@@ -1223,6 +1223,9 @@ Copilot will show completions only if all predicates return t."
                        (or
                         (string-prefix-p "copilot-" (symbol-name this-command))
                         (member this-command copilot-clear-overlay-ignore-commands)
+                        ;; `this-original-command' captures remapped helpers like
+                        ;; `universal-argument-more' and `digit-argument'.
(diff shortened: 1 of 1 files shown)


bbatsov:
That would be fine by me.

PR #436, opened 2025-12-09
Hardcode some commands that should not clear the overlay

ultronozm:
* copilot.el (copilot--hardcoded-clear-overlay-ignore-commands): New
constant.
(copilot--post-command): Use it.

### DIFF
--- a/copilot.el
+++ b/copilot.el
@@ -122,6 +122,10 @@ performance."
   :type '(repeat function)
   :package-version '(copilot . "0.1"))
 
+(defconst copilot--hardcoded-clear-overlay-ignore-commands
+  '(universal-argument digit-argument negative-argument universal-argument-more)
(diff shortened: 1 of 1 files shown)


copilot.el
```diff
@@ -122,6 +122,10 @@ performance."
   :type '(repeat function)
   :package-version '(copilot . "0.1"))
 
+(defconst copilot--hardcoded-clear-overlay-ignore-commands
```
bbatsov: Perhaps "default" is a better term than "hardcoded"?

ultronozm: Would you prefer something like this?

```elisp
(defconst copilot--clear-overlay-ignore-defaults
  '(universal-argument digit-argument negative-argument universal-argument-more)
  "Built-in list of commands that should not clear the overlay.")
```

PR #472, opened 2026-05-20
Support TRAMP by spawning remote servers

bastianbeischer:
Please note: The proposed changes in this PR were largely developed with the assitance of Copilot with Claude Sonnet 4.6 and Claude Opus 4.7...

- We now spawn one server per remote host (identified by tramp remote prefix)
- Use hash tables to track connections and servers

Also note: This also allows to use devcontainers where you launch the server within a restricted docker container and open files using tramp within the container. This allows sandboxing functionality of the copilot server, which many organisations require.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 ### New Features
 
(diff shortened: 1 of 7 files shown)


bastianbeischer:
- missing: user is currently expected to install the server themselves on the remote server.

Issue #477, opened 2026-06-16
Clarify relationship between copilot-mode and copilot-nes-mode


real-or-random:
### Description

The docs (README and docstring) say "copilot-nes-mode can coexist with copilot-mode". The wording "can" seems to indicate that copilot-nes-mode can also be without copilot-mode. But the truth is that copilot-nes-mode doesn't even start the copilot server without copilot.

### Steps to reproduce

1. Open a file 
2. Enable copilot-nes-mode without copilot-mode
3. Nothing happens, i.e., no NES suggestion and also not even an error message

### Emacs version

31.0.5 (2026-05-25)

### copilot.el version

99a6caf

### Copilot server version

1.476.0

### Node.js version

v26.2.0

### Error messages / logs

```text

```

### Operating system

Linux

real-or-random:
Also, I'm not sure if they should or can really coexist.

There seem to be issues when it happens that both modes suggest edits simultaneously.

Note: I don't use the suggested TAB key bindings for copilot-mode, so they don't interfere with the TAB binding from copilot-nes-mode. I have this instead: 
``` 
  :bind (:map copilot-completion-map
              ;; Inspired by fish
              ("<right>" . #'copilot-accept-completion)
              ;; C-(S-)<right> may not work in terminal but it's okay for me
              ("C-<right>" . #'copilot-accept-completion-by-word)
              ("C-S-<right>" . #'copilot-accept-completion-by-line)
              ("M-n" .  #'copilot-next-completion)
              ("M-p" . #'copilot-previous-completion))
```

### Example 1 

<img width="2678" height="232" alt="Image" src="https://github.com/user-attachments/assets/345c6ae9-75c3-4a0b-8746-f2cf2b6ca322" />

When I accept the copilot-mode suggestion (`<right>`), I get this: 

<img width="2724" height="278" alt="Image" src="https://github.com/user-attachments/assets/0cf9e695-9c7b-4d24-a42c-955b58b42131" />

This is not super useful. The NES suggestion should also disappear immediately. (Sorry, I haven't tried what happens if I press TAB now...

### Example 2

<img width="1122" height="334" alt="Image" src="https://github.com/user-attachments/assets/46ed3797-2af8-43e4-b89e-2479c1977c97" />

When I accept the copilot-nes-mode suggestion (`TAB`), I get this: 

[...]

Issue #451, opened 2026-02-27
NES insertion overlay is hard to see


bbatsov:
Reported by @dustinfarris in #447 — the deletion overlay (red/pink strikethrough) is visible, but the insertion text is easy to miss.

Two root causes:

1. **Face too subtle** — `copilot-nes-insertion-face` inherits `diff-added`, which in many themes only defines a faint background tint with default foreground. The text blends into normal buffer content.

2. **Multi-line rendering** — insertions use a single `after-string` overlay, so multi-line replacement text gets crammed at one position. Needs per-line overlays similar to how inline completions handle multi-line ghost text.

Ref: https://github.com/copilot-emacs/copilot.el/pull/447#issuecomment-3973369498

mwac-dev:
I have the same issue as @lucasec , insertions will just show up as a cross-out and then the line will get replaced, doesn't seem to be a theme issue.

unfortunately nothing in the logs stick out, 
the *copilot-language-server-log* will just show

[message 0 returned. finish reason: [stop]] 
[request done: requestId: [some-long-id] model deployment ID: []]

lucasec:
@bbatsov Sorry about the delay on this. I turned on the verbose logging and did manage to get this log. Oddly, not very much showed up in the events buffer.

[copilot-language-server-log.log](https://github.com/user-attachments/files/27000686/copilot-language-server-log.log)

bastianbeischer:
@bbatsov I debugged this and found out that the insertion overlay appears if I remove the evaporate property:

```
diff --git a/copilot-nes.el b/copilot-nes.el
index 1594124..109245c 100644
--- a/copilot-nes.el
+++ b/copilot-nes.el
@@ -187,7 +187,7 @@ Run `M-x copilot-reinstall-server' to upgrade."
                (ov (make-overlay end end nil nil nil)))
           (overlay-put ov 'after-string insertion-text)
           (overlay-put ov 'copilot-nes t)
-          (overlay-put ov 'evaporate t)
+          ;; (overlay-put ov 'evaporate t)
           (overlay-put ov 'priority 100)
           (push ov copilot-nes--overlays)))))
   ;; Record point so the post-command hook can detect actual movement
```

Maybe because `(make-overlay end end nil nil nil)` has `end end` it immediately evaporates itself?

EDIT: Yes I think that is the case:

evaporate
[...]

PR #478, opened 2026-06-17
Improve NES coexistence with copilot-mode

bbatsov:
NES relies on `copilot-mode` to start and sync the language server, but that wasn't enforced or documented, and the two modes stepped on each other's overlays. Addresses the points raised in #477 (and the keybinding question from the #451 thread).

- Dismiss a pending NES suggestion as soon as the buffer is edited (now including the `copilot-accept-completion*` commands), so accepting an inline completion no longer leaves a stale, misplaced NES overlay behind.
- Clear the NES overlays before applying an accepted suggestion, so the ghost text doesn't linger over the edit until the next command. Note: I couldn't reproduce the evil-mode redisplay glitch in Example 2 locally, so this is a best-effort fix for the most likely cause (the lingering after-string overlay) — worth confirming with the reporter.
- Warn when `copilot-nes-mode` is enabled without `copilot-mode`, and document the dependency in the docstring and README.
- Document why NES predefines TAB/C-g while `copilot-mode` leaves accept keys to `copilot-completion-map`.

I left true standalone NES (driving the doc lifecycle without `copilot-mode`) out of scope; this clarifies the current relationship instead.

Refs #477

### DIFF
--- a/README.md
+++ b/README.md
@@ -253,15 +253,20 @@ NES predicts the next edit you'll want to make anywhere in the file, based on yo
 >
 > NES requires `copilot-language-server` version 1.434.0 or newer. Run `M-x copilot-reinstall-server` to upgrade if needed.
 
-Enable `copilot-nes-mode` in a buffer to start receiving suggestions. It can coexist with `copilot-mode`:
+`copilot-nes-mode` works alongside `copilot-mode` and relies on it to start and sync the language server. Enable both in the buffer; on its own `copilot-nes-mode` does not produce any suggestions (it will tell you so when enabled without `copilot-mode`).
 
 ```elisp
+(add-hook 'prog-mode-hook #'copilot-mode)
 (add-hook 'prog-mode-hook #'copilot-nes-mode)
 ```
 
 When a suggestion is pending:
 - **TAB** — accept the suggestion (jumps to it first if far away, applies on second press)
 - **C-g** — dismiss the suggestion
+
+> [!NOTE]
+>
+> `copilot-nes-mode` predefines **TAB** and **C-g**, whereas `copilot-mode` ships no completion keybindings and leaves `copilot-completion-map` for you to populate. In both cases the bindings only take effect while a suggestion (or completion) is pending and otherwise fall through to their usual commands, so the predefined NES keys are safe to leave enabled.
 
 Customization variables:
 - **`copilot-nes-idle-delay`** — seconds of idle time before requesting a suggestion (default `0.5`)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,9 @@
 
 ### Bug Fixes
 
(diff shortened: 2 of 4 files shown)


lucasec:
All of this looks sensible to me. It should help with some of the conflicts I just saw in my testing, although we could probably go a step further and let NES shadow any regular copilot suggestions (this is what VSCode seems to do, but it might also be nice to give the user an option).

I'm a little torn on whether NES really makes sense as a separate mode. Given these are tightly-coupled (and the dependency of NES on the regular `copilot-mode`), NES may ought to just be a submode or toggle within the regular `copilot-mode`. Personally, once we get this really working well, I'm inclined to enable NES in all the same places I enable `copilot-mode`. I would have no complaints having one less entry on my modeline. Potential precedent for this would be the "electric mode" and "hungry delete" toggles in the C/C++ major modes, which are not their own minor modes but simply toggles within the major mode.

README.md
```diff
@@ -253,16 +253,21 @@ NES predicts the next edit you'll want to make anywhere in the file, based on yo
...
 (add-hook 'prog-mode-hook #'copilot-nes-mode)
 ```
 
 When a suggestion is pending:
 - **TAB** — accept the suggestion (jumps to it first if far away, applies on second press)
 - **C-g** — dismiss the suggestion
 
+> [!NOTE]
+>
+> Unlike `copilot-mode`, which ships no completion keybindings and leaves `copilot-completion-map` for you to populate, `copilot-nes-mode` predefines **TAB** and **C-g**. These are bound conditionally and only act while a suggestion is pending, falling through to their usual commands otherwise, so they are safe to leave on.
```
lucasec: `copilot-completion-map` also only acts when a suggestion is pending, no?

bbatsov: Good catch, you're right, both are pending-only. Reworded so the only real difference is that NES ships the bindings while `copilot-mode` leaves them to you.

bbatsov:
Thanks for taking a look. Agreed on shadowing, that's the cleaner fix for the both-fire-at-once case and matches what you saw in VSCode. I'll do it as a follow-up with an opt-out (`copilot-nes-shadow-completions`, on by default) so this PR stays focused on the overlay/coexistence fixes.

On folding NES into `copilot-mode` as a toggle rather than its own minor mode: I'd rather keep it separate, it's already released and merging it in would break existing `copilot-nes-mode` setups. The extra modeline entry is fair though, I'm happy to shorten the lighter.

lucasec:
Sounds good, thanks for the openness to feedback on all of this.

I'll have to think about how to configure my bindings for NES mode. I currently use a slightly more intentional sequence `C-c RET` for accepting regular copilot suggestions, as I don't like accidentally activating a suggestion when I habitually press TAB to indent. I assume I could bind the same key for NES mode, just a little cumbersome given the binding object with the filter property.

PR #480, opened 2026-06-17
Shadow copilot-mode completions while a NES suggestion is pending

bbatsov:
Follow-up to #478. When a regular `copilot-mode` completion and a NES suggestion want to show at the same time, they overlap on screen and accepting one can garble the other (see #477). This mirrors what the editor does in VSCode: prefer the Next Edit Suggestion and hide the inline completion while it's pending.

- New `copilot-nes--shadow-completion-p` disable-display predicate keeps `copilot-mode` from showing a completion while a NES edit is pending.
- Displaying a NES suggestion clears any completion already on screen.

Per the discussion in #477, there's no opt-out: offering two kinds of completion at once is overwhelming, and nobody has really run NES in production yet, so NES always wins when both would fire.

Refs #477

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 ### New Features
 
+- Hide regular `copilot-mode` completions while a NES suggestion is pending, so the two no longer overlap on screen. ([#477](https://github.com/copilot-emacs/copilot.el/issues/477))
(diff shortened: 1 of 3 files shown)


bbatsov:
On the timing edge real-or-random raised: a normal completion showing and then getting shadowed ~0.5s later while you're mid-decision is a real rough edge that this PR doesn't try to solve. Shadowing fixes the overlap/garble case, but the "suggestion yanked out from under you" scenario needs more thought (debounce? don't shadow once you've started interacting with the completion?). Tracking it separately rather than holding this up.

PR #496, opened 2026-06-25
Surface public code citations

### LINKED ISSUE #471: UI support for "Public Code References"
### Description

Copilot can detect if it suggests code that is very close to public code:
https://docs.github.com/en/enterprise-cloud@latest/copilot/how-tos/get-code-suggestions/find-matching-code

After trying the `fizzBuzz` example mentioned in the docs above, I can see this indeed in the Copilot log buffer.
```
[...]
[Public Code References] Text found matching public code in file:///[...].js [Ln 7, Col 13] near function fizzBuzz() {     for (let i = 1; i <= 100; i++) {         if (i % 3 === 0 && i % 5 === 0) {             console.log("FizzBuzz");         }         else if (i % 3 === 0) {             console.log("Fizz...:
  1) [GPL-2.0] https://github.com/github-copilot/code_referencing?cursor=57d70cbe624fc0cc762f9808e9ed375b,9e72d7433c6d28dd8a6b16bfe4529d1b,a6b458eec4cfae179b3bb252f5014d11,
  2) [MIT] https://github.com/github-copilot/code_referencing?cursor=57d70cbe624fc0cc762f9808e9ed375b,9e72d7433c6d28dd8a6b16bfe4529d1b,a6b458eec4cfae179b3bb252f5014d11,
  3) [unknown] https://github.com/github-copilot/code_referencing?cursor=57d70cbe624fc0cc762f9808e9ed375b,9e72d7433c6d28dd8a6b16bfe4529d1b,a6b458eec4cfae179b3bb252f5014d11,
[...]
```

So logging works (as expected) but it would be very nice to have some other (optional) indicator in the UI when Copilot suggests public code, e.g., a `(message ` or something similar.

### Alternatives considered

_No response_

### Additional context

_No response_

bbatsov:
GitHub Copilot can detect when a suggestion closely matches public code, but the server only emitted these as log lines. This declares the `ipCodeCitation` capability (only when the feature is enabled) and handles the `copilot/ipCodeCitation` notification: it announces the match in the echo area and collects each one, with its licenses and reference URLs, in a `*copilot-code-references*` buffer reachable via `M-x copilot-list-code-citations`. Controlled by `copilot-show-code-citations` (default on).

Fixes #471

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 ### New Features
 
+- Announce when a suggestion matches public code, and collect the matches (with licenses and reference URLs) in a buffer shown by `copilot-list-code-citations`. Controlled by `copilot-show-code-citations` (default on). ([#471](https://github.com/copilot-emacs/copilot.el/issues/471))
 - Track Copilot usage quota: surface the server's quota warnings and add `copilot-quota` to show how much of your chat, completion, and premium-request allowance is left.
 - Run the agent-mode `run_in_terminal` tool asynchronously instead of blocking Emacs for the whole command. The editor stays responsive, the language-server connection keeps flowing, `C-g` aborts a running command, and `copilot-chat-terminal-timeout` (default 30s) kills runaways. The command's exit status is now reported back to the model.
 - Show a status line in the chat buffer when Copilot runs one of its own built-in or MCP tools that asks for confirmation (`read_file`, the edit tools, etc.), so those tool calls leave a trace, not just the ones copilot.el executes locally.
--- a/README.md
+++ b/README.md
@@ -403,6 +403,7 @@ For example:
 | `copilot-logout` | Log out from GitHub Copilot |
 | `copilot-diagnose` | Restart the server and show diagnostic info |
 | `copilot-quota` | Show the current Copilot usage quota |
+| `copilot-list-code-citations` | Show suggestions that matched public code |
 | `copilot-select-completion-model` | Choose which model to use for completions |
 | `copilot-chat-select-model` | Choose which model to use for chat |
 | **Completion** | |
--- a/copilot.el
+++ b/copilot.el
@@ -137,6 +137,16 @@ find indentation offset."
   :type 'boolean
   :package-version '(copilot . "0.1"))
 
+(defcustom copilot-show-code-citations t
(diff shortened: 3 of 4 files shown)


real-or-random:
Thanks for implementing this. It works for me. Except that I don't see the message in the echo area. (Does this work for you? I only see it in the Messages buffer. This may be a problem with my config.)

I think the naming consistency here can be improved, and it would probably be nice to do this before the next release. Sometimes the code matches are called "citations" and sometimes "references". This is pretty confusing, e.g., if you set `copilot-show-code-citations`, then you get them in a buffer `*copilot-code-references*`. 

I suggesting calling them "references" consistently, because this is the official naming by GitHub:
 - https://docs.github.com/en/enterprise-cloud@latest/copilot/how-tos/get-code-suggestions/find-matching-code
 - https://github.com/github-copilot/code_referencing?cursor=840cf6e6594c948431b077a035c54e4c

bbatsov:
Thanks for the thoughtful feedback, Tim!

Good catch on the naming. You're right that "references" is GitHub's official term, so I've standardized on it everywhere -- `copilot-show-code-references` and `copilot-list-code-references`, with obsolete aliases for the old names so nobody's config breaks. That's in #505. I also added a README note clarifying these are GitHub's public-code references and not Emacs `xref`, since the overlap is easy to trip over.

On the echo-area message: I don't think it's your config. The notification fires while you're actively typing (it rides along with the inline completion), so the message gets clobbered by the next redisplay almost immediately -- it sticks around in `*Messages*` but only flashes by in the echo area. The `*copilot-code-references*` buffer is the reliable record. Keeping a message visible during active editing is genuinely tricky, but I'm open to ideas if you'd prefer something more persistent (a mode-line indicator, say).

real-or-random:
> The `*copilot-code-references*` buffer is the reliable record. Keeping a message visible during active editing is genuinely tricky, but I'm open to ideas if you'd prefer something more persistent (a mode-line indicator, say).

A simple thing will be an option (maybe off by default) that opens the buffer as a side window whenever something is added there.

A more sophisticated idea is to have a visual indicator as part of the completion itself, e.g., a different face than the one normally used for suggestions. (But I don't know if this makes sense timing-wise, i.e., whether the reference information comes in along with the code suggestion or later.)

PR #513, opened 2026-07-03
Add commit message generation from the staged diff

bbatsov:
Adds `copilot-chat-insert-commit-message`: run it in a Magit/`git-commit` buffer (or anywhere in a repo) and it generates a commit message from the staged diff and inserts it at point. This is the most-loved feature of chep's copilot-chat.el, adapted to the language-server architecture.

Implementation notes:

- The request runs as a throwaway one-shot conversation, so it neither needs the chat panel open nor disturbs an ongoing conversation there. To support that, `copilot-chat--active-buffers` entries can now hold a function sink besides a chat buffer; the buffer streaming path is unchanged, just factored into `copilot-chat--handle-buffer-progress`.
- The one-shot conversation is destroyed after the reply arrives, including the race where the reply completes before `conversation/create` returns the conversation id.
- The instruction is customizable via `copilot-chat-commit-message-prompt` (defaults to a conventional 72-char-summary format). Code fences are stripped defensively if the model adds them anyway.
- The server is auto-started if needed (same as other requests), so this works straight from a commit buffer in a fresh session.

425 specs pass, checkdoc clean, no new compile warnings.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@
   - `copilot-chat-doc`
   - `copilot-chat-optimize`
   - `copilot-chat-write-tests`
+- Add `copilot-chat-insert-commit-message` to generate a commit message from the staged changes and insert it at point (handy in a Magit or `git-commit` buffer), with the instruction sent to Copilot customizable via `copilot-chat-commit-message-prompt`.
 
 ### Bug Fixes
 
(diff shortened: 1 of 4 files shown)


bbatsov:
Addressed the adversarial-review findings: git now runs via `process-file` in `default-directory` (TRAMP buffers diff the remote repo; linked worktrees resolve their own index), function sinks are unregistered before their callback runs and only registered after the request was issued, callbacks are guarded against double invocation, `copilot-chat-stop` cancels a pending one-shot instead of resetting the chat conversation, a read-only target buffer gets the message on the kill ring instead of a process-filter error, four-backtick fences are stripped, and request errors render through `copilot-chat--error-text`. The staged-diff tests now run real git in temp repositories.

PR #516, opened 2026-07-03
Add a transient menu for the most common commands

bbatsov:
Adds `M-x copilot-menu`, a magit-style transient collecting the package's most common commands in one discoverable place: completions (toggle/complete/panel), chat (open, send region, slash commands, context, model selection), agent mode (toggle plus MCP tool listing), account/usage (login, logout, quota, code references), and server management (install, reinstall, uninstall, diagnose, log).

The toggles show live state inline (`Automatic completions [on]`, `Agent mode [off]`, `Select chat model (gpt-5)`), reading only variables already in memory, so opening the menu never triggers server requests.

`transient` is a soft dependency: current releases require Emacs 28.1 (where it's bundled anyway), so a hard dependency would break the package's 27.2 floor. The menu definition is stored as data and evaluated only when transient loads; the file byte-compiles fine without it and `copilot-menu` explains what's missing if invoked on such a setup.

New file `copilot-menu.el` with its own test file; 415 specs pass, checkdoc clean, compiles without warnings.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@
   - `copilot-chat-optimize`
   - `copilot-chat-write-tests`
 - Add `copilot-chat-insert-commit-message` to generate a commit message from the staged changes and insert it at point (handy in a Magit or `git-commit` buffer), with the instruction sent to Copilot customizable via `copilot-chat-commit-message-prompt` and a pending generation cancellable with `copilot-chat-stop`.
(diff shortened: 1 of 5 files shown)


bbatsov:
Reworked after the CI failures: current transient releases require Emacs 28.1, so a hard dependency would break the 27.2 floor. transient is now a soft dependency: the prefix definition is stored as data and evaluated only when transient loads (so the file byte-compiles on 27.2 without it), and `copilot-menu` signals a clear user-error when it's missing. The layout test now walks our own definition instead of transient's internal layout, which changed shape on Emacs snapshot.

PR #531, opened 2026-07-05
Let the copilot panel accept a solution

bbatsov:
`copilot-panel-complete` pops a `*copilot-panel*` buffer listing several suggestions for point, sorted by score, but there was no way to act on one: you had to read a solution and retype it into your buffer by hand.

This adds `copilot-panel-accept-completion`, bound to `RET` and `C-c C-c` in the panel via a small `copilot-panel-mode`. Move to the solution you want and press the key; the code block for that entry is inserted back at the exact position `copilot-panel-complete` was invoked from (tracked with a marker, so edits elsewhere in the source buffer meanwhile are fine), and the panel is dismissed.

The panel buffer is now read-only so `RET`/`C-c C-c` aren't swallowed by `self-insert`/`org-return`; the two solution inserters bind `inhibit-read-only` so the async `PanelSolution`/`PanelSolutionsDone` notifications can still populate it.

Tests cover extraction (heading and in-block), the nil case, insertion at the marker, the missing-origin error, and the keymap. README and CHANGELOG updated.


bbatsov:
Reworked after adversarial review. The first cut recovered the solution text by re-parsing the rendered Org source block, which meant a completion whose own content contained a `#+END_SRC` line (silent truncation) or a `*`-heading line (accept refused outright) round-tripped wrong. Common in Markdown/Org/prose/comment completions. Now the raw text is stored in a SHA-keyed table when the solution is rendered and recovered by the heading's SHA, so the block contents are never parsed for recovery. Added regression tests for both cases plus multi-line solutions.

PR #532, opened 2026-07-05
Add copilot-chat-compact to reclaim conversation context

bbatsov:
The language server exposes a `conversation/compress` method (verified in the 1.517.0 bundle: registered handler, params `{conversationId}`, result `{success, turnCount, contextInfo, ...}`). It summarizes the conversation so far and reclaims token budget while the conversation continues.

`copilot-chat-compact` wires it up (also in the `copilot-menu` chat section):
- Guards on an active, non-streaming conversation and a live connection.
- Sends `conversation/compress` with the active conversation id and a long timeout (the summarization is an LLM call; the 10s jsonrpc default would drop it).
- On success reports the turn count and resulting context usage; when the server declines (no new turns since the last compaction) it says so without erroring.

Worth knowing: the server already auto-compacts as the context fills (hard cap, idle, token limit). This is just an on-demand trigger of the same partition-compression machinery (`reason: "manual"`), so it coexists rather than fights it. The visible transcript is untouched; only the server-side context the next message is built from is compressed.

Tests cover the percentage formatter, both guards, the request shape, and the success / declined message paths. README and CHANGELOG updated.


bbatsov:
Fixed after review: the menu entry used key `c`, which collides with the existing `c` (Complete at point) in the same transient prefix (one flat keymap), silently shadowing it. Moved Compact to `C`. Added a menu-test assertion that every suffix key is distinct so this can't regress.

PR #533, opened 2026-07-05
Unlock policy-locked chat models from copilot-chat-select-model

bbatsov:
The `copilot/models` response already tells us which models are gated: each entry can carry `policy: {state, terms}` (verified in the 1.517.0 bundle; locked when `state` != `"enabled"`) and `billing: {is_premium, multiplier, ...}`. And the server exposes a real `copilot/setModelPolicy {model, status: "enabled"}` request (registered handler; it POSTs the acceptance to GitHub's backend and refetches metadata).

This uses both:
- `copilot-chat-select-model` now flags **premium** and **policy-locked** models in the completion list, e.g. `Claude Opus 4.8 (claude-opus-4-8) [premium, locked]`.
- Selecting a policy-locked model prompts with its `terms` and, on consent, sends `copilot/setModelPolicy` to accept them on the server. The model is switched only once the server confirms; declining leaves the current model untouched.

Before this, a policy-locked model could be selected but would just error on the next message, with no in-Emacs way to accept the terms.

Tests cover the annotation for plain/premium/locked/both, the unlocked path, the locked-consent path (asserts the request shape and that the model is set only from the success callback), and the declined path (nothing sent, model unchanged). README and CHANGELOG updated.


bbatsov:
Fixed after review: the unlock request was issued from whatever buffer M-x ran in, and `copilot--async-request` drops its success handler if that buffer is killed before the reply. So killing the origin buffer within the request window would let the server accept the policy while `copilot-chat-model` never updated. Now the request is issued from the persistent ` *copilot-chat-one-shot*` buffer (same precedent as the one-shot conversation requests), and there's a regression test that kills the origin buffer and asserts the unlock still records. Also documented that a malformed/stateless policy is treated as locked (errs toward asking for consent).

PR #534, opened 2026-07-05
Use native git/commitGenerate for commit messages

bbatsov:
The language server has a dedicated `git/commitGenerate` method (verified in the 1.517.0 bundle: registered handler, params `{changes, userCommits, recentCommits, workspaceFolder?, userLanguage?}`, result `{commitMessage}`). Unlike our one-shot chat, it feeds the model your recent commit subjects, so the generated message matches your repository's style, and it picks up the repository's commit instructions via the workspace folder.

This reworks `copilot-chat-insert-commit-message` to call it:
- `changes` is the staged diff, `userCommits`/`recentCommits` are recent commit subjects (yours and the repo's, for style reference), and `workspaceFolder` is the repo root URI.
- The reply comes back already stripped of code fences by the server, so the insertion logic is shared and the fence-stripping only runs on the fallback path.
- On a server too old to expose the method (method-not-found, -32601), it falls back to the previous one-shot conversation, so nobody loses the feature and `copilot-chat-commit-message-prompt` still applies there.

The request is issued from the always-live ` *copilot-chat-one-shot*` buffer (the callbacks target the commit buffer explicitly), matching the existing one-shot precedent so an aborted commit buffer can't drop the reply.

Tests cover the params builder (with and without a workspace root / user email), method-not-found detection, the shared insertion helper (insert / error / empty / read-only kill-ring), the native send-and-insert path, the fallback on method-not-found, and that other errors are reported without falling back. README and CHANGELOG updated.


bbatsov:
Tweaked after review: build the `workspaceFolder` URI with the canonical `copilot--path-to-uri` instead of a raw `file://` concat, so it percent-encodes paths with spaces/non-ASCII and produces the correct `file:///C:/...` form on Windows. Only affects whether server-side commit instructions load, but no reason to get the URI wrong.

PR #535, opened 2026-07-06
Route chat to BYOK models from the model picker

bbatsov:
First of two PRs adding Bring Your Own Key (BYOK) support (built-in providers: Anthropic, OpenAI, Gemini, Groq, OpenRouter, Azure). This one is the chat integration; the management commands (add/list/remove keys and models) follow in a second PR.

Verified against the 1.517.0 bundle: BYOK models never appear in `copilot/models` (that handler maps CAPI metadata only). They come from `copilot/byok/listModels`, and the chat path uses one by sending `modelInfo.providerName` on `conversation/create`/`conversation/turn` (the `resolveTurnModelInfo` server code reads `modelInfo.providerName ?? modelProviderName`). copilot.el previously sent `{modelInfo:{id}, model}` with no provider, so a BYOK model would have been treated as a CAPI id and failed to resolve.

Changes:
- `copilot-chat-select-model` now also lists registered BYOK models (from `copilot/byok/listModels`, best-effort so an old server without the method doesn't break selection), tagged `[BYOK: PROVIDER]`.
- Selecting a BYOK model tracks its provider in `copilot-chat--model-provider` alongside `copilot-chat-model`; selecting a Copilot model clears it.
- `copilot-chat--model-param` attaches `modelInfo.providerName` **only** for an explicit BYOK selection (guarded on `copilot-chat-model` being set), so a server-resolved default can never pick up a stale provider.

Note: using a BYOK model needs a BYOK-eligible account (the server gates the turn on a `client_byok` token; not something the client can advertise around). Storing/listing keys is ungated, which the follow-up PR builds on.

Tests: model-param attaches/omits providerName correctly (incl. the stale-default guard), byok-models filters to registered and degrades to nil on error, byok-choice formatting, and the picker's BYOK-select and switch-back-clears-provider paths. README and CHANGELOG updated.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@
 - Add `copilot-panel-accept-completion` to insert a solution from the `*copilot-panel*` buffer back into the buffer that requested it. Move to a suggestion and press `RET` or `C-c C-c`; previously the panel could only be read, never accepted from.
 - Add `copilot-chat-compact` (in the `copilot-menu` chat section) to compact the current chat conversation on demand: it asks the server to summarize the discussion so far, reclaiming token budget while the conversation continues. The visible transcript is left as-is, and it reports how many turns were compacted along with the resulting context usage.
 - `copilot-chat-select-model` now flags premium and policy-locked models in the completion list, and selecting a policy-locked one prompts to accept its terms and enables it on the server (via `copilot/setModelPolicy`) before switching, instead of leaving you to hit an error on the next message.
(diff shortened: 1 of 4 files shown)


bbatsov:
Self-caught while building the follow-up: `copilot-chat-apply-preset` set `copilot-chat-model` directly without clearing `copilot-chat--model-provider`, so applying a preset with a Copilot model while a BYOK provider was still active would have sent a mismatched `providerName`. Fixed (presets carry Copilot model ids in this MVP, so the provider is cleared), with a regression test.

PR #536, opened 2026-07-06
Add BYOK key and model management commands

bbatsov:
Second of two BYOK PRs. Stacked on #535 (the chat integration); this adds the commands to register and manage what the picker surfaces.

- `copilot-chat-byok-add-key` -> `copilot/byok/saveApiKey`. The key is read with `read-passwd` (no echo) and handed to the language server, which stores it. copilot.el never keeps or displays a key, and never calls `listApiKeys` (which echoes keys back in plaintext). Azure keys are per-model, so it asks for a model id.
- `copilot-chat-byok-add-model` -> `copilot/byok/saveModel` (id, display name, tool-call/vision capabilities; Azure also needs a deployment URL).
- `copilot-chat-byok-list`, `copilot-chat-byok-remove-model` (`deleteModel`), `copilot-chat-byok-remove-key` (`deleteApiKey`, with confirmation).

All schemas verified against the 1.517.0 bundle. Requests go through a small shared helper that issues from the always-live ` *copilot-chat-one-shot*` buffer (so a killed origin buffer can't drop the reply) and reports the server's own success/error message.

Security posture: keys are write-only from the client's side; nothing is logged, displayed, or read back. The plaintext-on-disk storage is the server's (`byok.json`, mode 0600), out of our hands.

Tests cover each command's request shape (including Azure's modelId/deploymentUrl), the empty-key guard, the no-models and declined-confirmation paths, list output, and the shared helper preferring the server message. README and CHANGELOG updated (the BYOK entry now covers the full add -> use flow).

Note: depends on #535; will rebase onto main once that lands.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,7 +16,11 @@
 - Add `copilot-panel-accept-completion` to insert a solution from the `*copilot-panel*` buffer back into the buffer that requested it. Move to a suggestion and press `RET` or `C-c C-c`; previously the panel could only be read, never accepted from.
 - Add `copilot-chat-compact` (in the `copilot-menu` chat section) to compact the current chat conversation on demand: it asks the server to summarize the discussion so far, reclaiming token budget while the conversation continues. The visible transcript is left as-is, and it reports how many turns were compacted along with the resulting context usage.
 - `copilot-chat-select-model` now flags premium and policy-locked models in the completion list, and selecting a policy-locked one prompts to accept its terms and enables it on the server (via `copilot/setModelPolicy`) before switching, instead of leaving you to hit an error on the next message.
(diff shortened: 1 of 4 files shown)


bbatsov:
Folded in the review's non-blocking notes: add-model now validates the model id right after reading it (fail fast, before the capability prompts); remove-key omits Azure from the provider list (its keys are per-model, so deleteApiKey always rejected it) and points to remove-model instead; and the README now warns that turning on `copilot-log-max` would persist a BYOK key in the `*copilot events*` buffer. Review confirmed the `:json-false` booleans encode correctly and no key leaks through copilot.el (read-passwd, never stored/logged/displayed, listApiKeys never called).

PR #537, opened 2026-07-06
Support custom OpenAI-compatible endpoints for BYOK

bbatsov:
BYOK phase 2: custom (non-built-in) OpenAI-compatible endpoints, on top of the built-in providers from #535/#536.

Verified against the 1.517.0 bundle: `copilot/byok/saveCustomProviderConfig {providerName, apiKey, groupName, apiType?}` stores the key AND the config in one call (`apiType` is one of `chatCompletions`/`responses`/`messages`), and rejects built-in provider names. A custom provider's models are registered with `saveModel` carrying a `deploymentUrl` (required, same as Azure), and the provider must be configured first. Custom-endpoint models route to chat the same way as any BYOK model (`modelInfo.providerName`), which #535 already wired up.

- `copilot-chat-byok-add-custom-provider`: prompts for a custom name (validated not to be a built-in), API type, group name (defaults to the provider name), and key (read-passwd). One `saveCustomProviderConfig` call.
- `copilot-chat-byok-add-model`: the provider list now includes your custom providers; Azure and custom providers require a per-model deployment URL, and a custom-endpoint model is saved with `isCustomModel` set (user-defined, not a catalog model the provider advertises). Built-in non-Azure providers are unchanged.
- `copilot-chat-byok-remove-key`: offers custom providers too (`deleteApiKey` removes their config and models; Azure still omitted).
- `copilot-chat-byok-list-custom-providers`: lists configured custom providers with their API type.

Same security posture as #536: keys read with `read-passwd`, never stored/logged/displayed, and no new call to `listApiKeys`.

Tests cover the custom/built-in predicate, listing custom providers (and error -> nil), add-custom-provider's request shape and built-in-name rejection, add-model's custom-provider path (deploymentUrl + isCustomModel), the empty-URL guard, remove-key including custom providers, and list-custom-providers output. README and CHANGELOG updated (the BYOK entry now covers custom endpoints).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@
 - `copilot-chat-select-model` now flags premium and policy-locked models in the completion list, and selecting a policy-locked one prompts to accept its terms and enables it on the server (via `copilot/setModelPolicy`) before switching, instead of leaving you to hit an error on the next message.
 - Add Bring Your Own Key (BYOK) support, letting you use your own model-provider API keys (Anthropic, OpenAI, Gemini, Groq, OpenRouter, Azure) in chat:
   - `copilot-chat-byok-add-key` saves a provider API key (read without echoing and held by the language server, never stored or shown by copilot.el), and `copilot-chat-byok-add-model` registers a model to use with it.
+  - `copilot-chat-byok-add-custom-provider` registers a custom OpenAI-compatible endpoint (any provider name that is not built in, with its API type and key); `copilot-chat-byok-add-model` then registers models for it (each with its deployment URL), and `copilot-chat-byok-list-custom-providers` shows the ones you have configured.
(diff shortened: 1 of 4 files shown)


bbatsov:
Review came back with one real (cosmetic) defect, now fixed: `copilot-chat-byok-list-custom-providers` wraps a synchronous request whose `jsonrpc-error` data isn't the shape `copilot-chat--error-text` expects, so an error printed a raw sexp dump. It now pulls the server message out of the jsonrpc-error alist (with `error-message-string` as a fallback), with a test. Everything else the review probed was refuted against the actual handlers: `isCustomModel:true` is inert metadata (stored, never validated, doesn't affect chat resolution), `deleteApiKey` does delete the custom provider config (no orphan), and case-variant provider names agree between client and server. Security clean.

